PR #11, opened 2019-01-15
D-50 retrieve and display user details in the global header

lbennett-moj:


assets/sass/application.scss
```diff
@@ -96,3 +97,22 @@ $path: "/assets/images/";
   margin-bottom:15px;
 }
 
+div.govuk-width-container {
```
lbennett-moj: a few tweaks - we can revisit when discussing styles

PR #22, opened 2019-01-25
DC-51 Display Cat A alert on 'Offending history' page

steverendell:


server/routes/form.js
```diff
@@ -42,9 +42,11 @@ module.exports = function Index({ formService, offendersService, authenticationM
       const pageData = getIn([section, form], res.locals.formObject)
       const errors = req.flash('errors')
       const details = await offendersService.getOffenderDetails(res.locals.user.token, bookingId)
+      // TODO needs rearranging into dedicated router call
```
lbennett-moj: yes - maybe we need to decide the best way of doing this before it gets too messy.

PR #38, opened 2019-02-08
DC-79 Display a list of offenders that have been categorised BUT need…

steverendell:
… reviewing in a particular prison


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/EscapeSpecification.groovy
```diff
@@ -41,9 +40,11 @@ class EscapeSpecification extends GebReportingSpec {
     when: 'I go to the tasklist page'
 
     elite2api.stubUncategorised()
```
steverendell: I simplified the sorting to not need the reverse() but this caused some problems with the tests, so i forced
the dates to be different in the stubs

package.json
```diff
@@ -47,6 +47,7 @@
     "joi": "^14.3.0",
     "joi-date-extensions": "^1.2.0",
     "joi-postcode": "^3.0.0",
+    "jwt-decode": "^2.2.0",
```
steverendell: I realised you can just pull the user's roles out of the JWT rather than making an api call as NN does

server/routes/home.js
```diff
@@ -8,6 +9,25 @@ module.exports = function Index({ authenticationMiddleware, userService, offende
 
   router.get(
     '/',
```
steverendell: I did a redirect here to allow for users who may have both roles and want to see either list page at some point in the near future!

server/services/offendersService.js
```diff
@@ -88,7 +88,53 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
...
+      const offenders = decoratedResults.sort((a, b) => sortByDateTimeDesc(a.dateRequired, b.dateRequired))
+
+      return offenders
+    } catch (error) {
+      logger.error(error, 'Error during getUncategorisedOffenders')
+      throw error
+    }
+  }
+
+  async function getUnapprovedOffenders(token, agencyId) {
```
steverendell: this might need some tidying, extracting common code

test/services/offenderService.test.js
```diff
@@ -132,7 +132,7 @@ describe('getUncategorisedOffenders', () => {
     moment.now.mockReturnValue(moment('2019-01-16', 'YYYY-MM-DD'))
     expect(service.buildSentenceData('2019-01-14')).toEqual({
       daysSinceSentence: 2,
```
steverendell: this looks a bit odd - dateRequired gets displayed but sentenceDate doesnt.

PR #61, opened 2019-02-27
Dc 131 refactor form.js

steverendell:


server/routes/form.js
```diff
@@ -193,7 +168,13 @@ module.exports = function Index({
     res.locals.formId = formData.id
 
     const backLink = req.get('Referrer')
-    const pageData = firstItem(req.flash('userInput')) || getIn([section, form], res.locals.formObject)
+
```
steverendell: this awkward bit reconstructs user input under the right section/form

lbennett-moj: yeah - at least it only has to go in one place i spose

PR #74, opened 2019-03-11
DC_137 set status. Further work to remove unnecessary status setting …

lbennett-moj:
…functions to come

### DIFF
--- a/integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/model/DatabaseUtils.groovy
+++ b/integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/model/DatabaseUtils.groovy
@@ -31,6 +31,11 @@ class DatabaseUtils {
     sql.executeUpdate("insert into form values (-1, ?::JSON, $bookingId, 'CATEGORISER_USER', 'STARTED', 'CATEGORISER_USER')", json)
   }
 
+  def createDataWithStatus(bookingId, status, json) {
+    def sql = Sql.newInstance(dbConnParams)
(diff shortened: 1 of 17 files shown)


server/utils/statusEnum.js
```diff
@@ -8,9 +8,10 @@ const states = {
   APPROVED: { name: 'APPROVED', value: 'Approved' },
 }
 states.SECURITY_MANUAL.previous = [states.STARTED, states.SECURITY_AUTO, states.SECURITY_BACK]
-states.SECURITY_AUTO.previous = [states.STARTED]
+states.SECURITY_AUTO.previous = [undefined, states.STARTED]
```
lbennett-moj: can remove started from the previous section when auto trigger is moved in later ticket

integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/model/DatabaseUtils.groovy
```diff
@@ -31,6 +31,11 @@ class DatabaseUtils {
     sql.executeUpdate("insert into form values (-1, ?::JSON, $bookingId, 'CATEGORISER_USER', 'STARTED', 'CATEGORISER_USER')", json)
   }
 
+  def createDataWithStatus(bookingId, status, json) {
+    def sql = Sql.newInstance(dbConnParams)
```
steverendell: $status in single quotes .. does that work?

lbennett-moj: it was the only way i could get it to work -  looks fine on the db etc

PR #82, opened 2019-03-15
don't lock button for any status

lbennett-moj:


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/SecurityInputSpecification.groovy
```diff
@@ -89,16 +88,8 @@ class SecurityInputSpecification extends GebReportingSpec {
     def today = LocalDate.now().format('DD/MM/YYYY')
     $('#securitySection').text().contains("Manually referred to Security ($today)")
 
-    when: 'I view the categoriser list again'
```
steverendell: perhaps its still worth checking the 'Manually referred to Security' status ?

PR #94, opened 2019-03-23
Dc 140 Allow categorisers and supervisors to view approved categorisations from their 'done' tab

steverendell:


assets/sass/application.sass
```diff
@@ -58,14 +58,48 @@ $path: "/assets/images/"
     width: 75%
```
steverendell: As part of this PR, Ive removed the duplication of sass files, opting for the indented version seeing as this is used as part of the build process. It was slightly out of date compared to the scss file though

server/services/formService.js
```diff
@@ -96,7 +96,7 @@ module.exports = function createFormService(formClient) {
     }
     if (
       (data.history && data.history.catAType) ||
-      (data.securityBack && data.securityBack.catB === 'Yes') ||
```
steverendell: slight bug fix!

server/views/formPages/approvedView.html
```diff
@@ -0,0 +1,100 @@
...
+
+<div class="govuk-grid-row govuk-body-s">
+  <div class="govuk-grid-column-two-thirds">
+
+    <h1 class="govuk-heading-l">Categorisation outcome</h1>
+
+    {% set cat = data.categoriser.provisionalCategory.suggestedCategory %}
+    {% set overriddenCat = data.categoriser.provisionalCategory.overriddenCategory %}
+    {% set categoriserCat = overriddenCat or cat %}
+    {% set superCat = data.supervisor.review.supervisorOverriddenCategory if data.supervisor.review.supervisorCategoryAppropriate == 'No' %}
```
lbennett-moj: I thought we were having a common content section then separate top level pages for each role?

PR #97, opened 2019-03-26
Dc 209 comment box

lbennett-moj:


server/views/formPages/approvedView.html
```diff
@@ -79,16 +79,16 @@ <h2 class="govuk-heading-s">Supervisor decision</h2>
...
-     {{ govukSummaryList({
-        classes: 'govuk-!-margin-bottom-9',
-        rows: [
-          {
-            key: { text: "Other relevant information" },
-            value: { text: data.tobedetermined },
-            actions: { items: [] }
-          }
-        ]
-      }) }}
```
steverendell: Im not too keen on losing indentations .. ideally we need a nunjucks formatter but for now i normally do 'paste without formatting' to avoid this

PR #105, opened 2019-04-01
DC-225 Remove 'sent back to categoriser' records from supervisor home list

steverendell:


server/services/offendersService.js
```diff
@@ -145,31 +144,44 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
...
-              displayName: `${properCaseName(o.lastName)}, ${properCaseName(o.firstName)}`,
-              categoriserDisplayName: `${properCaseName(o.categoriserFirstName)} ${properCaseName(
-                o.categoriserLastName
-              )}`,
-              dbRecordExists: !!dbRecord.booking_id,
-              ...buildSentenceData(sentenceMap.find(s => s.bookingId === o.bookingId).sentenceDate),
-            }
-          })
+        unapprovedOffenders.map(async o => {
+          // TODO: async probably now redundant ^^
```
JamieDraperUK: You are correct. You no longer need the async or the promise all. Just the unapprovedOffenders.map(o => would do it

server/services/offendersService.js
```diff
@@ -145,31 +144,44 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
...
       const nomisClient = nomisClientBuilder(token)
       const uncategorisedResult = (await nomisClient.getUncategorisedOffenders(agencyId)).filter(
-        s => s.status === Status.AWAITING_APPROVAL.name
+        s => s.status === Status.AWAITING_APPROVAL.name // the status coming back from nomis
       )
 
-      if (isNilOrEmpty(uncategorisedResult)) {
+      const unapprovedWithDbRecord = await Promise.all(
+        uncategorisedResult.map(async s => {
+          const dbRecord = await formService.getCategorisationRecord(s.bookingId)
```
JamieDraperUK: Very minor but to make it consistent with line 125 you change this to:

`
uncategorisedResult.map(async s => ({
    ...s,
   ...(await formService.getCategorisationRecord(s.bookingId)),
})
`

steverendell: ok il go for it !

server/services/offendersService.js
steverendell: ok ill go for it !

server/services/offendersService.js
steverendell: think ill leave this one: the code creates a field called dbrecord and is the same as at line 111

PR #130, opened 2019-04-18
dc-258 if supervisor overrides with a cat D or J - send back to categ…

lbennett-moj:
…oriser

### DIFF
--- a/server/routes/form.js
+++ b/server/routes/form.js
@@ -409,26 +409,31 @@ module.exports = function Index({
       const section = 'supervisor'
       const form = 'review'
       const formPageConfig = formConfig[section][form]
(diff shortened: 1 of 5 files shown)


server/routes/form.js
```diff
@@ -409,26 +409,31 @@ module.exports = function Index({
       const section = 'supervisor'
       const form = 'review'
       const formPageConfig = formConfig[section][form]
-
-      if (!formService.isValid(formPageConfig, req, res, section, form, bookingId)) {
-        return
+      if (req.body.supervisorOverriddenCategory !== 'D' && req.body.supervisorOverriddenCategory !== 'J') {
+        if (!formService.isValid(formPageConfig, req, res, section, form, bookingId)) {
```
lbennett-moj: dont need to validate if open conditions input is required as we aren't persisting the supervisor form

PR #139, opened 2019-04-29
security home page is using database to drive the display of referrals

lbennett-moj:


server/data/dataAccess/db.js
```diff
@@ -1,9 +1,11 @@
 const logger = require('../../../log')
 const config = require('../../config')
 
-const { Pool } = require('pg')
+const { Pool, types } = require('pg')
 const fs = require('fs')
 
+types.setTypeParser(20, val => parseInt(val, 10))
+
```
lbennett-moj: this seems to  be the recommended way to get pg to auto convert bigint (just booking Id currently) to an int.
I think we are save to do this and it means we don't have to have parseInts everywhere when comparing the db result

server/data/formClient.js
```diff
@@ -7,7 +7,7 @@ const sequenceClause =
 module.exports = {
   getFormDataForUser(bookingId) {
     const query = {
-      text: `select id, booking_id, user_id, status, form_response, assigned_user_id, referred_date, referred_by, security_reviewed_date, security_reviewed_by
+      text: `select id, booking_id as "bookingId", user_id, status, form_response, assigned_user_id, referred_date, referred_by, security_reviewed_date, security_reviewed_by
```
lbennett-moj: I think that removing the database column names from the code would be an improvement. If you agree - I'll change the other ones

test/services/offenderService.test.js
```diff
@@ -215,9 +221,72 @@ describe('getUncategorisedOffenders', () => {
   })
 })
```
lbennett-moj: had to move a few things as there was some strange nesting of tests

PR #158, opened 2019-05-14
limit done tab diplay (approved categorisations) by a configurable AP…

lbennett-moj:
…PROVED_DISPLAY_MONTHS env var. Required new approval_date column


server/services/formService.js
```diff
@@ -39,6 +40,31 @@ module.exports = function createFormService(formClient) {
     throw new Error(`Invalid state transition from ${currentCategorisation.status} to ${status}`)
   }
```
steverendell: i suppose there should be a test for this

PR #160, opened 2019-05-15
Dc 306 recat tasklist

steverendell:


server/routes/tasklistRecat.js
```diff
@@ -0,0 +1,72 @@
...
+        transactionalDbClient
+      )
+      // If retrieved - check if APPROVED and if it is, create new
+      if (categorisationRecord.status === Status.APPROVED.name) {
+        categorisationRecord = await formService.createCategorisationRecord(
+          bookingId,
+          req.user.username,
+          details.agencyId,
+          details.offenderNo,
+          categorisationRecord.sequence + 1,
```
lbennett-moj: should we be exposing the sequence number back to the router?  Would it be better to get the formClient to handle the increment?

server/services/offendersService.js
```diff
@@ -456,6 +456,23 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
     return result
   }
 
+  async function isRecat(token, bookingId, transactionalDbClient) {
+    // Decide whether initial or recat.
+    // If there is a local record then if that is APPROVED it is a recat, otherwise initial (continue existing)
+    // If no local record, get cat from nomis. if missing or UXZ it is initial, otherwise recat.
+    const localRecord = formService.getCategorisationRecord(bookingId, transactionalDbClient)
+    if (localRecord.status) {
+      return localRecord.status === Status.APPROVED.name
```
lbennett-moj: what if recat inprogress?

PR #179, opened 2019-06-05
Dc 323 add content to task list page

steverendell:
Also an ie8 fix and recat todo list fix


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/recat/HigherSecurityReviewSpecification.groovy
```diff
@@ -36,11 +36,12 @@ class HigherSecurityReviewSpecification extends GebReportingSpec {
     when: 'I go to the Higher Security Review page'
     fixture.gotoTasklistRecat(false)
     at TasklistRecatPage
-    higherSecurityReviewButton.click()
```
steverendell: temp fix to test until cat page comes along

server/views/partials/offenderDetails.html
```diff
@@ -29,58 +34,58 @@
 
     {{ headerTitle('Nationality') }}
```
steverendell: rest in this file is just restoring indentation

PR #192, opened 2019-06-14
CAT-286 Build check your answers page (recats)

steverendell:


server/services/formService.js
```diff
@@ -295,6 +295,25 @@ module.exports = function createFormService(formClient) {
     }
   }
```
steverendell: We could merge setAwaitingApproval and backToCategoriser to a general purpose status change function maybe at some point

server/routes/recat.js
```diff
@@ -243,6 +245,48 @@ module.exports = function Index({ formService, offendersService, userService, au
...
+            token: res.locals.user.token,
+            bookingId,
+            suggestedCategory,
+            overriddenCategoryText: 'Cat-tool Recat',
+            nextReviewDate,
+          })
+
+          const nextPath = getPathFor({ data: req.body, config: formPageConfig })
+          res.redirect(`${nextPath}${bookingId}`)
+        } else {
```
steverendell: Actually im not sure about this - maybe the open conditions flag should be set by the decision page

PR #196, opened 2019-06-21
cat-266 supervisor approval (recat)

lbennett-moj:


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/pages/SupervisorRecatReviewPage.groovy
```diff
@@ -0,0 +1,54 @@
+package uk.gov.justice.digital.hmpps.cattool.pages
```
steverendell: Would it be ok to put this class in the recat package?

PR #206, opened 2019-06-26
Cat 254 Offender showing on both uncategorised and supervisor todo lists

steverendell:


server/views/pages/categoriserHome.html
```diff
@@ -62,8 +62,9 @@ <h1 class="govuk-heading-l">Prisoner Categorisation</h1>
         {% set buttonText = "Edit" if row.dbRecordExists else "Start" %}
         {% set basicClasses = "tableButton govuk-!-padding-left-3 govuk-!-padding-right-3" %}
         {% set classes = basicClasses + " grey-button" if row.dbRecordExists else basicClasses %}
+        {% set inconsistent = row.status === 'AWAITING_APPROVAL' and (row.dbStatus === 'STARTED' or row.dbStatus === 'APPROVED') %}
```
lbennett-moj: Could we determine the inconsistency in the route to allow logging of the inconsistent data? It would be good to use the logs to work out how many we are getting or even alert the prison to this behaviour.

steverendell:
Ah yes I forgot about that

server/views/pages/categoriserHome.html
steverendell: 👍  moved logic into service and added logging

PR #231, opened 2019-07-12
Cat 413 cat history snapshot

lbennett-moj:


server/routes/form.js
```diff
@@ -513,6 +513,20 @@ module.exports = function Index({
         })
         await offendersService.createSupervisorApproval(res.locals.user.token, bookingId, userInput)
```
steverendell: could this go inside the if statement?

PR #241, opened 2019-07-18
Helm deploy migrate

mattops:


helm_deploy/values-prod.yaml
```diff
@@ -0,0 +1,23 @@
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
```
Mjwillis: replicaCount should be 2

mattops: updated

Mjwillis: 👍

PR #245, opened 2019-07-19
Cat_425 new agency agnostic endpoint for u21 lookups

lbennett-moj:
overriding page limit for initial u21 query
added flag to prevent 404 when prison is inactive

### DIFF
--- a/server/services/offendersService.js
+++ b/server/services/offendersService.js
@@ -314,7 +314,7 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
         }
         if (dbRecordExists && row.dbRecord.nomisSeq !== row.assessmentSeq) {
           logger.warn(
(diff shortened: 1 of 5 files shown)


server/services/offendersService.js
```diff
@@ -434,16 +434,19 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
   }
 
   async function mergeU21ResultWithNomisCategorisationData(nomisClient, agencyId, resultsU21IJ) {
-    const eliteCategorisationResultsU21 = await nomisClient.getLatestCategorisationForOffenders(
+    const eliteResultsRaw = await nomisClient.getLatestCategorisationForOffenders(
       agencyId,
-      resultsU21IJ.map(c => c.bookingId)
+      resultsU21IJ.map(c => c.offenderNo)
     )
```
steverendell: Im a bit worried about when the latest cat is status I, it looks like you are then filtering the prisoner from the recat todo list .. is that right?. Id think that logically we actually want the latest A or P , but I could be wrong. Probably Nic/Louise could tell us when this can happen and how to interpret it?

PR #249, opened 2019-07-23
Getting problems in production opening the security screens when an o…

lbennett-moj:
…ffender has transferred to another agency. Changed to another endpoint to return offender names that doesn't limit to current agency


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/mockapis/Elite2Api.groovy
```diff
@@ -718,23 +718,23 @@ class Elite2Api extends WireMockRule {
   }
 
 
-  def stubGetOffenderDetailsByBookingIdList(String agency) {
+  def stubGetOffenderDetailsByOffenderNoList(List offenderNumbers) {
```
lbennett-moj: I'll change from list to 2 distinct offenderno parameters in next PR

PR #257, opened 2019-08-01
CAT-437 Message from supervisor

steverendell:


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/pages/SupervisorMessagePage.groovy
```diff
@@ -0,0 +1,23 @@
...
+class SupervisorMessagePage extends Page {
+
+  static url = '/form/supervisor/supervisorMessage'
+
+  static at = {
+    headingText == 'Message from supervisor'
+  }
+
+  static content = {
+    headingText { $('h1.govuk-heading-l').text() }
```
petergphillips: not sure about all this selecting elements based on govuk styles, makes the tests very brittle if govuk change...

server/views/formPages/supervisor/confirmBack.html
```diff
@@ -1,38 +1,58 @@
...
+    id: "messageText",
+    maxlength: 50000,
+    threshold: 50,
+    label: {
+      text: "Please enter a message for the categorisor"
+    },
+    value: data.supervisor.confirmBack.messageText,
+    errorMessage: {
+      text: "Please enter details"
+    } if (errors | findError('transferText'))
```
petergphillips: should this be `messageText` instead of `transferText`?

PR #262, opened 2019-08-05
CAT-446 Overridden sentence dates are not being taken into account

steverendell:


server/views/partials/offenderDetails.html
```diff
@@ -61,28 +61,28 @@
...
     {% if data.details.sentence.paroleEligibilityDate %}
       {{ headerTitle('Parole Eligibility Date') }}
       {{ headerValue(dateConverter(data.details.sentence.paroleEligibilityDate)) }}
     {% endif %}
-    {% if data.details.sentence.nonParoleDate %}
+    {% if data.details.sentence.nonParoleOverrideDate or data.details.sentence.nonParoleDate %}
       {{ headerTitle('Non Parole Date') }}
-      {{ headerValue(dateConverter(data.details.sentence.nonParoleDate)) }}
+      {{ headerValue(dateConverter(data.details.sentence.nonParoleOverrideDate or data.details.sentence.nonParoleDate)) }}
     {% endif %}
```
Robc24: LGTM.  Is there any way of setting **data.details.sentence** as it's own variable and using something more concise rather than **data.details.sentence.ETC** every time?

steverendell: good idea

Robc24: Awesome.

PR #288, opened 2019-09-03
CAT-508 Cat A warning bug

steverendell:
Fixed assuming a cat A could not occur during the current sentence, and also applied ordering (as endpoint cat history data is actually not ordered! )


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/ratings/OffendingHistorySpecification.groovy
```diff
@@ -68,7 +68,6 @@ class OffendingHistorySpecification extends GebReportingSpec {
     db.getData(12).status == ["STARTED"]
   }
 
-
```
lbennett-moj: do we have some end to end test coverage on the display outcome? It feels like the end html should be tested

PR #289, opened 2019-09-03
continued risk profile alert and bug CAT-509

lbennett-moj:


server/utils/riskChange.js
```diff
@@ -0,0 +1,59 @@
+const { equals } = require('../utils/functionalHelpers')
```
lbennett-moj: all wip until requirements are nailed down

server/utils/riskChange.js
```diff
@@ -0,0 +1,59 @@
...
+    escapeRiskAlert,
+    escapeList,
+    escapeRisk,
+    alertRequired: false, // **disabled** escapeListAlert || escapeRiskAlert || escapeRisk || escapeList,
+  }
+}
+
+function alertCompare(a, b) {
+  const alertIdA = a.alertId
+  const alertIdB = b.alertId
```
steverendell: one trick i learned recently for integers is that you can just return alertIdA - alertIdB

lbennett-moj: noted!

PR #296, opened 2019-09-06
Cat 508 cat a warning bug 2

steverendell:


server/services/offendersService.js
```diff
@@ -624,12 +624,13 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
...
   }
 
-  async function getCatAInformation(token, offenderNo, currentBookingId) {
+  async function getCatAInformation(token, offenderNo, currentBookingIdString) {
     try {
       const nomisClient = nomisClientBuilder(token)
       const categories = await getCategoryHistoryWithoutPendingCategories(nomisClient, offenderNo)
       const sortedCategories = categories.sort(sortByDescendingBookingAndAscendingSequence)
       const mostRecentCatA = sortedCategories.find(isCatA)
+      const currentBookingId = parseInt(currentBookingIdString, 10)
```
steverendell: Problem was that I was comparing an integer with a string

PR #320, opened 2019-10-02
DT-249: updated session secret

SimonMitchellMOJ:


Dockerfile
```diff
@@ -1,8 +1,14 @@
 FROM node:10.15-slim
-MAINTAINER HMPPS Digital Studio <info@digital.justice.gov.uk>
+LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 ARG BUILD_NUMBER
 ARG GIT_REF
 
+RUN apt-get update && apt-get install -y make python && \
```
JamieDraperUK: Do we need 'make' and 'python'?

PR #323, opened 2019-10-04
Cat 553 refinement of dashboard stats

steverendell:


server/data/statsClient.js
```diff
@@ -34,13 +34,40 @@ module.exports = {
...
 
+  getRecatFromTo(startDate, endDate, prisonId, transactionalClient) {
+    const query = {
+      text: `
+        with cat_table as (
+          select booking_id,
+                 sequence_no,
+                 coalesce (form_response -> 'supervisor' ->'review' ->>'supervisorOverriddenCategory',
+                           form_response -> 'recat' -> 'decision' ->>'category') as cat
+          from form)
```
steverendell: I went with the temp table solution in the end , i.e. 'cat_table ' but hopefully it is not actually all that complex. Its like a subselect in the from clause in oracle

server/views/pages/dashboardRecat.html
```diff
@@ -24,8 +24,31 @@ <h1 class="govuk-heading-l">Dashboard - Categorisation reviews</h1>
 <div>
   {% include "../partials/dashboardForm.html" %}
   <hr/>
+  <h3>Categorisation review numbers</h3>
```
steverendell: a cycler is a nunjucks function which cycles round, returning each of its values in turn

PR #407, opened 2020-04-30
CAT-586 Add other categories to the service - approval pages

steverendell:


server/views/pages/liteApprove.html
```diff
@@ -0,0 +1,213 @@
...
+      attributes: {  'id': 'inProgressWarning' }
+    }) }}
+
+    <div>
+      <a href="/" id="todoLink1">Todo list</a>
+      <a href="/{{ bookingId }}" id="manageLink1">Manage categorisations</a>
+    </div>
+
+  {% elif sameUser %}
+
```
lbennett-moj, line 50: Do we enforce this for the main approvals?

steverendell: no but i put it in here simply because old nomis does it - i suppose users would be used to this

PR #409, opened 2020-05-01
CAT-636 UI tweaks for other categories

steverendell:
Suggestions from Marius

### DIFF
--- a/server/routes/liteCategories.js
+++ b/server/routes/liteCategories.js
@@ -3,7 +3,7 @@ const moment = require('moment')
 const flash = require('connect-flash')
 const baseJoi = require('joi')
 const dateExtend = require('joi-date-extensions')
-const { calculateNextReviewDate, sanitisePrisonName } = require('../utils/utils')
+const { properCaseName, calculateNextReviewDate, sanitisePrisonName } = require('../utils/utils')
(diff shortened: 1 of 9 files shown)


server/routes/liteCategories.js
```diff
@@ -12,9 +12,37 @@ const joi = baseJoi.extend(dateExtend)
...
+
+const getCatList = current => [
+  { value: 'U', text: 'Unsentenced', selected: current === 'U' },
+  { value: 'Z', text: 'Unclass', selected: current === 'Z' },
+  { value: 'A', text: 'Cat A', selected: current === 'A' },
+  { value: 'E', text: 'Cat A Ex', selected: current === 'E' },
+  { value: 'H', text: 'Cat A Hi', selected: current === 'H' },
+  { value: 'P', text: 'Prov A', selected: current === 'P' },
+  { value: 'R', text: 'Fem Closed', selected: current === 'R' },
+  { value: 'T', text: 'Fem Open', selected: current === 'T' },
```
lbennett-moj, line 43: Is this used anywhere else?  Just wondering if an enum would help

steverendell: I just refactored these from 2 copies of select values - one in each form html page. I could create an enum but this takes a parameter which makes it a little different?

PR #450, opened 2020-12-23
CAT-555 Flagging to security can block progress

steverendell:
Also CAT-651 Content Changes to the Recat process, item 4

### DIFF
--- a/server/routes/home.js
+++ b/server/routes/home.js
@@ -385,8 +385,20 @@ module.exports = function Index({
       const user = await userService.getUser(res.locals)
       const details = await offendersService.getOffenderDetails(res.locals, bookingId)
 
+      // ensure cat not in progress.
+      const assessmentData = await formService.getLiteCategorisation(bookingId, transactionalDbClient)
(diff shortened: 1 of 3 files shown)


server/routes/home.js
```diff
@@ -385,8 +385,20 @@ module.exports = function Index({
       const user = await userService.getUser(res.locals)
       const details = await offendersService.getOffenderDetails(res.locals, bookingId)
```
steverendell, line 387: We just need to block the race condition that one person starts a categorisation and another arrives here to "flag" the offender

server/views/formPages/openConditions/riskOfHarm.html
```diff
@@ -94,7 +94,7 @@ <h1 class="govuk-heading-l">{{ pageTitle }}</h1>
   {% set guidanceSeriousHarm %}
 <div class="govuk-!-font-weight-bold">You should check:</div>
 <ul class="govuk-list govuk-list--bullet">
-  <li>risk of serious harm (ROSH) score</li>
+  <li>OASys assessment</li>
```
steverendell, line 97: the Content Change

connormaglynn (approved):
LGTM

PR #451, opened 2020-12-24
CAT-678 Show supervisor name when rejected

steverendell:
Pass supervisor name back to page shown to the categoriser


connormaglynn (approved):
LGTM

PR #454, opened 2020-12-31
CAT-676 Cancelling a manual initial cat leaves all nomis cats inactive

steverendell:
Now includes in-progress cases from the database as well as Nomis


connormaglynn (approved):
LGTM

PR #455, opened 2021-01-04
Cat 679 flag pf cases to security 2

steverendell:
Minor tweak to content

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "passport": "^0.4.1",
     "passport-local": "^1.0.0",
     "passport-oauth2": "^1.5.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
     "passport": "^0.4.1",
     "passport-local": "^1.0.0",
     "passport-oauth2": "^1.5.0",
-    "pg": "^7.18.2",
+    "pg": "^8.0.3",
```
steverendell, line 69: update to fix problem with node 14.x

steverendell: backed this out as doesnt work with node 12

lbennett-moj (approved):

PR #460, opened 2021-01-27
CAT-686: Remove Lite assessments that are approved in NOMIS

connormaglynn:
On pressing 'approve' on the Approve page. We catch the error back from Nomis, delete the record then redirect the page below.

Think we can probably do something a bit smarter but thought I'd get the first draft up.

Couple of further changes?
- Adding in the details of the approval on the 'Already Approved' page
- Checking something to gauge whether this error will be thrown and either automatically delete the record or offer the option to the user (failing fast). Noticed that the history page for the nominal has the categorisation in there, getting that list might provide enough detail, or there might be something better.

image


docker-compose-test.yml
```diff
@@ -0,0 +1,26 @@
+version: '3.1'
```
connormaglynn, line 1: Happy to expand on this a bit more or scrap it?

server/routes/liteCategories.js
```diff
@@ -313,24 +317,44 @@ module.exports = function Index({ formService, offendersService, userService, au
...
+            nextReviewDate,
+            approvedPlacement,
+            approvedPlacementComment,
+            approvedComment,
+            transactionalClient: transactionalDbClient,
+          })
+          res.redirect(`/liteCategories/confirmed/${bookingId}`)
+        } catch (error) {
+          if (assessmentHasBeenApprovedInNomisMannually(error)) {
+            formService.deleteLiteCategorisation(bookingId, assessmentData.sequence, transactionalDbClient)
```
steverendell, line 338: missing await

server/routes/liteCategories.js
```diff
@@ -313,24 +317,44 @@ module.exports = function Index({ formService, offendersService, userService, au
...
       }
     })
   )
 
+  router.get(
+    '/alreadyApproved/:bookingId',
+    asyncMiddleware(async (req, res) => {
+      const { bookingId } = req.params
+      const user = await userService.getUser(res.locals)
+      res.locals.user = { ...user, ...res.locals.user }
```
steverendell, line 353: The endpoint getCategoryHistory() uses would return assessment details from nomis but sadly the approver is not included - it would need to be added in prison-api. probably not worth it atm

server/views/pages/liteAlreadyApproved.html
```diff
@@ -0,0 +1,34 @@
...
+
+{% include "../partials/breadCrumb.html" %}
+
+{% endblock %}
+
+{% block content %}
+
+  {{ govukPanel({
+    titleText: "Categorisation has already been approved"
+  }) }}
```
steverendell, line 18: I think the govukWarning (with that '!' icon would be better than the green panel - to show that what has happened is 'less than ideal' !

steverendell (approved):
lgtm :)

PR #468, opened 2021-04-19
Cat 689 unapproved cats lost when moving prison

steverendell:
Added movement event handler whcih updates 4 tables

### DIFF
--- a/server/services/offendersService.js
+++ b/server/services/offendersService.js
@@ -1337,6 +1337,48 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
     )
   }
 
+  const handleExternalMovementEvent = async (
+    context,
+    bookingId,
+    offenderNo,
+    movementType,
+    fromAgencyLocationId,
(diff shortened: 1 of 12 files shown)


server/services/offendersService.js
```diff
@@ -1337,6 +1337,60 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
...
+            const resultSecurity = await formService.updatePrisonSecurityReferral(
+              offenderNo,
+              toAgencyLocationId,
+              client
+            )
+            results.push({ name: 'riskChange', ...resultRiskChange })
+            results.push({ name: 'securityReferral', ...resultSecurity })
+          }
+          logger.info(
+            `Movement summary: rows updated =${results.reduce((s, item) => `${s} ${item.name}: ${item.rowCount}`, '')}`
```
steverendell, line 1382: This seemed a bit better than having 4 result vars which may or may not have been set

lbennett-moj (approved):

PR #470, opened 2021-04-21
Cat 687 merge event not updating all tables 2

steverendell:


server/services/formService.js
```diff
@@ -766,24 +766,63 @@ module.exports = function createFormService(formClient) {
...
+
+    const securityReferralRemove = await getSecurityReferral(oldOffenderNo, transactionalClient)
+    const securityReferralSurvives = await getSecurityReferral(newOffenderNo, transactionalClient)
+
+    // If row for surviving offenderNo exists, or to-be-removed offenderNo exists, use that.
+    // If both exist, use whichever is not in COMPLETED status
+    // If neither COMPLETED, use the later raisedDate
+    // If both COMPLETED, do nothing.
+    // NOTE no need to delete the removed record when not used.
+
```
steverendell, line 793: This is horrible because offenderNo is a unique key and its possible (though unlikely) that both the offenderNos have been flagged, with e.g. , the one to be deleted having the 'NEW' status, so being the record we want to keep.

connormaglynn (approved):
lgtm

PR #476, opened 2021-05-18
PAT-1782 upgrade cat tool to node 14 - QA fix

steverendell:
Fixed loss of bold headings and margin: govuk now sets `<p>` margins, and to be explicitly non-bold by default


package.json
```diff
@@ -84,7 +84,7 @@
     "jest": "^26.6.3",
     "nock": "^12.0.3",
     "prettier": "^2.2.1",
-    "redis-cli": "^1.4.0",
+    "redis-cli": "^2.0.0",
```
steverendell, line 87: vulnerability update

lbennett-moj (approved):

PR #480, opened 2021-06-01
PAT-1765 cancel security flag

steverendell:


docker-compose-test.yml
```diff
@@ -60,14 +60,8 @@ services:
     entrypoint: /bin/sh -c
     command: >
       "
-        while [[ $$(curl -s -o /dev/null -w ''%{http_code}'' localstack:4572) != "200" ]]; do
-          echo 'DCS AWS - awaiting localstack service start'
-          sleep 10
-        done
-
         aws --endpoint-url=http://localstack:4576 sqs create-queue --queue-name event
```
steverendell, line 63: Thisa check doesnt work for me, but doesnt seem to be needed anyway

package.json
```diff
@@ -37,7 +37,6 @@
     "body-parser": "^1.19.0",
     "bunyan": "^1.8.15",
     "bunyan-format": "^0.2.1",
-    "bunyan-request-logger": "^2.1.0",
```
steverendell, line 40: This module has some serious security probs now, and no fixed version. Think we can do without it as we can always use requests rather than traces (requests gives less info tho :( )

lbennett-moj (approved):

PR #481, opened 2021-06-14
PAT-1920 DCS logging broken

steverendell:


server/catToolSerialisers.js
```diff
@@ -0,0 +1,15 @@
+const bunyan = require('bunyan')
```
steverendell, line 1: file reinstated as it was

yasinmustafa (approved):
lgtm

PR #485, opened 2021-06-21
⚡️PAT-1778: add security tab for flagged nearing next review

connormaglynn:
image

image

image


server/services/offendersService.js
```diff
@@ -324,6 +324,68 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
     }
   }
 
+  async function getUpcomingReferredOffenders(context, transactionalDbClient) {
+    const agencyId = context.user.activeCaseLoad.caseLoadId
+    try {
+      const nomisClient = nomisClientBuilder(context)
+
+      const securityReferred = await formService.getSecurityReferrals(agencyId, transactionalDbClient)
```
connormaglynn, line 332: Did we want to restrict these to just ones that are 'NEW' and 'REFERRED'?

server/services/offendersService.js
```diff
@@ -775,6 +838,15 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
           const status = sortByStatus(b.dbStatus, a.dbStatus)
           return status === 0 ? sortByDateTime(b.nextReviewDateDisplay, a.nextReviewDateDisplay) : status
         })
+        .map(o => {
+          return {
+            ...o,
+            securityReferred:
+              securityReferredOffenders
+                .filter(s => s.offenderNo === o.offenderNo)
+                .filter(s => s.status === 'NEW' || s.status === 'REFERRED').length > 0,
```
connormaglynn, line 847: Restricted them for the label on the recat page, as I thought it didn't make much sense to display tag for a cancelled referral?

server/views/pages/recategoriserHome.html
```diff
@@ -64,7 +65,14 @@ <h1 class="govuk-heading-l">Category reviews for prisoners</h1>
           {{ row.nextReviewDateDisplay }}
         {% endif %}
       </td>
-      <td class="govuk-table__cell"><a target="_blank" href="{{ row.offenderNo | offenderLink }}">{{ row.displayName }}</a></td>
+      <td class="govuk-table__cell"><a target="_blank" href="{{ row.offenderNo | offenderLink }}">{{ row.displayName }}</a>
+        {% if row.securityReferred %}
+        {{ govukTag({
+        text: 'Security Referred',
```
connormaglynn, line 71: Happy to change this text if we think there's something better?

server/views/pages/securityUpcoming.html
```diff
@@ -0,0 +1,83 @@
...
+      <td class="govuk-table__cell"><a target="_blank" href="{{ row.offenderNo | offenderLink }}">{{ row.displayName }}</a></td>
+      <td class="govuk-table__cell">{{ row.offenderNo }}</td>
+      <td class="govuk-table__cell">{{ row.daysSinceSentence }}</td>
+      <td class="govuk-table__cell">{{ row.dateRequired }}</td>
+      <td class="govuk-table__cell">{{ row.securityReferredBy or 'Automatic' }}</td>
+      <td class="govuk-table__cell govuk-!-margin-top-2 govuk-!-margin-bottom-8">
+        {% set basicClasses = "tableButton govuk-!-padding-left-3 govuk-!-padding-right-3" %}
+        {% set classes = basicClasses + " grey-button" %}
+        {{ govukButton({
+          text: 'Manage',
```
connormaglynn, line 66: Again not sure if manage is the right word, but happy for other ideas!

server/services/offendersService.js
connormaglynn: Also did we want to display the **Status** on the final page instead of the **Days since sentence**?

steverendell: Its just new (as it is only for offenders whose categorisation hasnt started yet). So that means status is not relevant either 😄 .
And also , yes, I dont think 'Days since sentence' is worth showing either.

server/views/pages/recategoriserHome.html
steverendell: That looks good, but maybe some hover text or similar would be good too , e.g. 'The security team have arranged that they will automatically be referred to security when their next category review is started'. Possibly should involve Martin as Im not sure hover text is very 'accessible'.

server/views/pages/securityUpcoming.html
steverendell: be good to follow other existing pages, see cancellation or other confirmation pages

server/services/offendersService.js
```diff
@@ -324,6 +324,68 @@ module.exports = function createOffendersService(nomisClientBuilder, formService
...
+          let securityReferredBy
+          if (o.userId) {
+            const referrer = userDetailFromElite.find(record => record.username === o.userId)
+            securityReferredBy = referrer
+              ? `${properCaseName(referrer.lastName)}, ${properCaseName(referrer.firstName)}`
+              : o.userId
+          }
+
+          const entry = sentenceMap.get(offenderDetail.bookingId)
+          const sentenceAndDate = entry && buildSentenceData(entry.sentenceDate)
```
steverendell, line 368: You wont need sentence data if we dump that column 👍

server/services/offendersService.js
steverendell: maybe .some() is preferable here

server/services/offendersService.js
steverendell: On second thoughts, the security user can flag someone who requires an initial categorisation, (sorry I didnt think of this) .. so I think we will have to show the same 'Days since sentence' 'Date due' and 'type' columns as the security home page list for these,

server/views/pages/recategoriserHome.html
steverendell: Also As this can apply to initial cats too, we need a label on the initial todo list as well

server/services/offendersService.js
```diff
@@ -88,12 +88,19 @@ function calculateRecatDisplayStatus(displayStatus) {
   return displayStatus === Status.APPROVED.value || !displayStatus ? 'Not started' : displayStatus
 }
 
+function isSecurityReferred(offenderNo, securityReferredOffenders) {
+  return securityReferredOffenders
+    .filter(s => s.offenderNo === offenderNo)
+    .some(s => s.status === 'NEW' || s.status === 'REFERRED')
+}
```
steverendell, line 95: needs to be just 'NEW' - both for list and tags :)

steverendell (approved):
lgtm

PR #488, opened 2021-06-28
pat-1783 For a due recat, the review reason was being taken from a pr…

lbennett-moj:
…evious categorisation (if it exists), changed for >21s and standard recats


steverendell (approved):
lgtm :)

PR #509, opened 2021-11-08
PAT-2098 Convert to use generic-service

steverendell:


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ orbs:
 executors:
   builder:
     docker:
-      - image: cimg/node:14.17.6-browsers
+      - image: cimg/node:14.18.1-browsers
```
steverendell, line 10: sight upgrade while we are here

monskapinga (approved):

PR #536, opened 2022-06-22
DSE-310: Added serviceUnavaliable page for 503 page

thirlz95:


Burdis:
QA Reviewed.

steverendell:
@thirlz95 yes I believe that is correct 🙂

PR #540, opened 2022-07-22
CAT-959 amend ingress to enable parallel run.

kkonuganti:


thirlz95 (approved):

helm_deploy/offender-categorisation/values.yaml
```diff
@@ -10,11 +10,10 @@ generic-service:
...
   ingress:
     enabled: true
-    modsecurity_enabled: false
-    modsecurity_audit_enabled: false
-    tlsSecretName: "offender-categorisation-cert"
-    path: /
-    contextColour: green
+    modsecurity_enabled: true
+    v1_2_enabled: true
+    annotations:
```
alex-hedley, line 15: Is this needed?

kkonuganti: sorry dint see your comment. 
Yes v1_2_enabled will enable tls 1.2 on the application. IN common helm chart we already have v_0_47_enabled parameter as true by default. So did not add it here.

Common annotations already exist in generic template, but in order for them to be active we need ti add empty parameter, as per the template.
Modsec was not enabled before on this project, dont know why. That means there is no security or WAF security enabled. So i have enabled it now.

steverendell: Be careful about enabling modsecurity, it may cause things to get blocked which are not immediately apparent. Needs good regression test

PR #584, opened 2022-10-28
cat-729 amend breadcrumbs

kkonuganti:
- remove last link aria-current page
- remove change Categorisation home to Categorisation dashboard
- change Task list page heading to Categorisation task list
- amend tests

### DIFF
--- a/server/views/macros/breadCrumb.html
+++ b/server/views/macros/breadCrumb.html
@@ -1,6 +1,6 @@
 {% from "breadcrumbs/macro.njk" import govukBreadcrumbs %}
 
-{% macro breadCrumb(pageTitle, breadCrumbList) %}
+{% macro breadCrumb(breadCrumbList) %}
 
   {% set rows = [ {
     text: "Digital Prison Services",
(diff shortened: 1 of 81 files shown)


server/views/macros/breadCrumb.html
```diff
@@ -3,7 +3,7 @@
 {% macro breadCrumb(pageTitle, breadCrumbList) %}
```
kberec, line 3: pageTitle isn't used anymore

server/views/macros/breadCrumb.html
```diff
@@ -3,7 +3,7 @@
...
 
   {% set rows = [ {
-    text: "Home",
+    text: "Digital Prison Services",
     href: dpsHome
```
kberec, line 7: This can cause an a11y issue.

kberec (changes requested):

server/views/macros/breadCrumb.html
kberec: As discussed, it seems to be an existing issue, just check if there is some easy way to resolve it with these changes.

kberec (approved):

PR #592, opened 2022-11-15
Cat 1251 orb update

Olusola-Opencast:


kberec (approved):

kkonuganti (approved):
after looking at the circle ci orb conf, dont think we need to add anything else.

PR #596, opened 2022-11-21
cat-1318 update cron jobs

kkonuganti:
- update cron jobs to be cleared every 4 days.
- update cron job to use latest k8 version

### DIFF
--- a/database_backup/02-cronjob.yaml
+++ b/database_backup/02-cronjob.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: batch/v1beta1
+apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: rds-snapshot-job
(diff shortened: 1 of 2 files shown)


database_backup/02-cronjob.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: batch/v1beta1
+apiVersion: batch/v1
```
kkonuganti, line 2: this change is to pick up latest k8 version to be used.

database_backup/02-cronjob.yaml
```diff
@@ -10,6 +10,7 @@ spec:
   failedJobsHistoryLimit: 3
```
kkonuganti, line 10: this job is run 1st day of every, so setting the job to be cleaned after 4 days of running.

database_refresh/02-cronjob.yaml
```diff
@@ -10,6 +10,7 @@ spec:
   failedJobsHistoryLimit: 3
   jobTemplate:
```
kkonuganti, line 11: this batch is run every sunday. So 4 days to clean the job should still be viable.

Olusola-Opencast (approved):
ok

kberec (approved):

PR #597, opened 2022-11-22
modified to show recat note

Olusola-Opencast:


server/views/formPages/recat/securityBack.html
```diff
@@ -14,14 +14,18 @@
...
 
-  {% set warningText %}
-    This prisoner was referred to the security team
-  {% endset %}
-  {{ govukWarningText({
-    text: warningText,
-    iconFallbackText: "Warning",
-    classes: "forms-warning-text warning-overrides"
-  }) }}
+  {% if (not data.socProfile.transferToSecurity) and (not data.extremismProfile.notifyRegionalCTLead) %}
```
kkonuganti, line 17: i think i note is for both initial and recat.

kkonuganti (changes requested):
There are two security back pages, one for CAT and one for RECAT.
For Recat we should just display the text.
For initial we need follow the design on the ACs

server/views/formPages/recat/securityBack.html
Olusola-Opencast: thanks

kberec (approved):

server/views/formPages/recat/securityBack.html
```diff
@@ -14,14 +14,17 @@
...
 
-  {% set warningText %}
-    This prisoner was referred to the security team
-  {% endset %}
-  {{ govukWarningText({
-    text: warningText,
-    iconFallbackText: "Warning",
-    classes: "forms-warning-text warning-overrides"
-  }) }}
+  {% if data.recat.securityInput.securityInputNeededText %}
```
kkonuganti, line 17: i think the condition is still wrong.
SecurityInputNeededText will be populated only when you say Yes to securityNoteNeeded.
So it might be like this
if(SecurityNoteNeeded == No){
    Note Was added
}elseif securityInputNeededText{
    show securityInputNeededText.
}

in terms of auto and flagged scenarios for security neither securityNoteNeeded nor securityInputNeededText will be populated.

kkonuganti (changes requested):

server/views/formPages/recat/securityBack.html
Olusola-Opencast: Just tested - the string in the database 
{"recat": {"oasysInput": {"date": "24/09/2019", "oasysRelevantInfo": "No"}, **"securityInput": {"securityInputNeeded": "Yes", "securityInputNeededText": "This kind of competence only comes from practice."}**, "prisonerBackground": {"offenceDetails": "no"}}, **"security": {"review": {"securityReview": "Furious and Riddick movies these days, so a brand-new franchise would help him out."}}**}
Allows the change to be backwards compatible with existing cases that don't have the **securityNoteNeeded** variable

Olusola-Opencast: Also modified two cases with:
-"securityInput": {"securityNoteNeeded": "", "securityInputNeeded": "Yes"}
- "securityInput": {"securityNoteNeeded": "", "securityInputNeeded": ""} 
And it works

kkonuganti: how about when there is no security input at all for RECAT as during flagged and auto there will be no 
"securityInput": {"securityNoteNeeded": "", "securityInputNeeded": ""}
can we also test if status is AUTO and there is no securityInput element in json at all

Olusola-Opencast: These scenarios have already been tested.

Olusola-Opencast: **transferToSecurity - true**
{"recat": {"securityBack": {}}, "security": {"review": {"securityReview": "Despite the violence and chaos and angry debate in Parliament, the measure passed easily, mostly because the ministers who were opposed either resigned or were forced out of the country's two ruling coalition parties."}}}

{"socProfile": {"nomsId": "G2843GF", "riskType": "SOC", "transferToSecurity": true, "provisionalCategorisation": "C"}, "extremismProfile": {"nomsId": "G2843GF", "riskType": "EXTREMISM", "notifyRegionalCTLead": false, "increasedRiskOfExtremism": false, "provisionalCategorisation": "C"}}

kkonuganti: LGTM

kkonuganti (approved):

PR #602, opened 2022-11-30
cat-1196 add oasys integration tests

kkonuganti:


Olusola-Opencast (commented):
To ensure there are no issues with the dates being invalid for example if this code runs in 2030 becoming in valid you can do something like this: 
  static final VALID_OASYS_DATE = LocalDate.now().minusMonths(1).format('dd/MM/yyyy')
  static final INVALID_OASYS_DATE = LocalDate.now().plusYears(6).format('dd/MM/yyyy')

integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/recat/OasysInputSpecification.groovy
```diff
@@ -0,0 +1,142 @@
...
+    submitButton.click()
+
+    then: "error messages should be displayed"
+    then: 'I stay on the page with validation errors'
+
+    at OasysInputPage
+    errorSummaries[0].text() == 'Completion date of the latest full review must be a real date'
+    errorSummaries[1].text() == 'Select yes if there was any information in the review that is relevant to the recategorisation'
+
+    when: 'Not a correct data is entered and Yes or No is not clicked'
```
kberec, line 107: Wouldn't it be better to test this as a unit test (e.g. test/config/openConditions.test.js) and only test here that an (any) error message is shown?

kberec (changes requested):

kberec (approved):

Olusola-Opencast (approved):
ok

PR #609, opened 2022-12-05
SDI-314 Change search api in the categorisation tool to elastic search

steverendell:
Moved some endpoints so cat tool now calls prisoner search (Elastic search ES) rather than prison-api as this is much more efficient and lightens the Nomis database load


server/data/nomisClientBuilder.js
```diff
@@ -102,10 +92,6 @@ module.exports = context => {
       const path = `${apiUrl}api/bookings/${bookingId}/mainOffence`
       return nomisClientGet({ path })
     },
-    getMainOffences(bookingIds) {
-      const path = `${apiUrl}api/bookings/mainOffence`
-      return nomisClientPost({ path, body: bookingIds })
-    },
```
steverendell, line 108: These are the 3 endpoint calls that have bee moved to prisoner search, getSentenceDatesForOffenders and getMainOffences both now use the same prisoner search endpoint as the data from both is returned from it

server/services/offendersService.js
```diff
@@ -32,50 +32,34 @@ function isOverdue(dbDate) {
...
   return new Map(
-    sentenceDates
-      .filter(s => s.sentenceDetail.sentenceStartDate) // the endpoint returns records for offenders without sentences
-      .map(s => {
-        const { sentenceDetail } = s
-        return [sentenceDetail.bookingId, { sentenceDate: sentenceDetail.sentenceStartDate }]
-      })
+    prisoners
+      .filter(s => s.sentenceStartDate) // the endpoint returns records for offenders without sentences
+      .map(s => [s.bookingId, { sentenceDate: s.sentenceStartDate }])
```
steverendell, line 45: I could have gone further and referred to sentenceStartDate in calls raher than sentenceDate

server/services/offendersService.js
```diff
@@ -144,19 +135,31 @@ module.exports = function createOffendersService(nomisClientBuilder, allocationC
...
       ])
 
+      const prisonerMap = new Map(prisoners.map(prisoner => [prisoner.bookingId, prisoner]))
+
+      const sentenceMap = new Map(
+        prisoners
+          .filter(s => s.sentenceStartDate) // the endpoint returns records for offenders without sentences
+          .map(s => [s.bookingId, { sentenceDate: s.sentenceStartDate }])
+      )
+
```
steverendell, line 156: getPrisonersByBookingIds() gets the data for both maps. In fact 2 separate maps are not really needed either

server/services/offendersService.js
```diff
@@ -144,19 +135,31 @@ module.exports = function createOffendersService(nomisClientBuilder, allocationC
...
       const filterIS91s = o => {
-        const offence = offenceMap.get(o.bookingId)
+        const offence = prisonerMap.get(o.bookingId)
         if (!offence) {
           return true
         }
-        if (offence.offenceCode === 'IA99000-001N' && offence.statuteCode === 'ZZ') {
+        if (offence.mostSeriousOffence === 'ILLEGAL IMMIGRANT/DETAINEE') {
```
steverendell, line 162: Description is in ES, not the codes

server/services/offendersService.js
```diff
@@ -906,7 +924,9 @@ module.exports = function createOffendersService(nomisClientBuilder, allocationC
   }
 
   async function mergeU21ResultWithNomisCategorisationData(nomisClient, agencyId, resultsU21IJ) {
-    const eliteResultsRaw = await nomisClient.getLatestCategorisationForOffenders(resultsU21IJ.map(c => c.offenderNo))
+    const eliteResultsRaw = await nomisClient.getLatestCategorisationForOffenders(
+      resultsU21IJ.map(c => c.prisonerNumber)
```
steverendell, line 928: ES data uses prisonerNumber rather than offenderNo

integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/mockapis/Elite2Api.groovy
```diff
@@ -323,35 +291,6 @@ class Elite2Api extends WireMockRule {
...
-                age         : 20,
-                categoryCode: 'I',
-              ],
-              [
-                bookingId   : 22,
-                offenderNo  : 'C0002AA',
-                firstName   : 'ADRIAN',
-                lastName    : 'MOLE',
-                dateOfBirth : '1998-08-15',
-                age         : 20,
```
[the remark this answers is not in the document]
steverendell: yes ✔️

server/services/offendersService.js
steverendell: There is a small risk, but there has always been a risk that a different code could be used too. Actually it is worth QA testing this case: of one of the prisoners being an IS91

nnehabharti: @steverendell While checking on T3 Nomis, we have found that the Offence Code and Status Code is still valid.Is it worthwhile if we keep using the same instead of description.
Screenshot 2022-12-16 at 14 33 23

server/data/prisonerSearchApi.js
```diff
@@ -26,14 +26,38 @@ module.exports = context => {
...
     async getPrisonersAtLocation(agencyId, fromDob, toDob) {
       const path = `${apiUrl}prison/${agencyId}/prisoners`
-      const query = `fromDob=${fromDob}&toDob=${toDob}`
-      return apiGet({ path, query })
+      // use size to effectively turn off paging
+      const query = `size=1000000&fromDob=${fromDob}&toDob=${toDob}`
+      const response = await apiGet({ path, query })
+      return response.content.map(r => ({
+        ...r,
+        bookingId: Number(r.bookingId),
```
steverendell, line 34: booking ids come back as string - convert to number for compatibility

server/services/offendersService.js
steverendell: Hi @nnehabharti , unfortunately I had to change this because the codes are not available in the data from the new service, only the description is present.

kkonuganti (approved):

PR #611, opened 2022-12-07
cat-1323 further charges cya changes

kkonuganti:


server/routes/openConditions.js
```diff
@@ -26,29 +26,28 @@ module.exports = function Index({ formService, offendersService, userService, au
     asyncMiddleware(async (req, res, transactionalDbClient) => {
       const { bookingId } = req.params
       const form = 'furtherCharges'
-      const result = await buildFormData(res, req, 'openConditions', form, bookingId, transactionalDbClient)
+      let result = await buildFormData(res, req, 'openConditions', form, bookingId, transactionalDbClient)
```
kberec, line 29: Generally in js it is preferred to have immutable objects, probs that's why this code was using the newResult variable further down the line. Making it mutable simplifies the code though, so I think the change make sense.

server/views/formPages/openConditions/furtherCharges.html
```diff
@@ -18,7 +18,7 @@
 {% endblock %}
 
 {% set furtherChargesHtml %}
-  {% if catType == 'INITIAL' %}
+  {% if catType == 'INITIAL'  and data.ratings.furtherCharges.furtherCharges == 'Yes' %}
```
kberec, line 21: Is it nullsafe? Can any elements in any scenarios in`data.ratings.furtherCharges` be null/undefined? 
Same question applies to line 66 below as well.

kberec: it was tested

kberec (approved):

server/views/formPages/openConditions/furtherCharges.html
Olusola-Opencast: Do you mean 
{% if catType == 'INITIAL'  and data.ratings.furtherCharges.furtherCharges **===** 'Yes' %}
rather than 
{% if catType == 'INITIAL'  and data.ratings.furtherCharges.furtherCharges **==** 'Yes' %}
?

kkonuganti: nope we will not do a strict check, we are going to just use loose check, going along with catType == 'INITIAL'

Olusola-Opencast (approved):
ok

PR #615, opened 2022-12-09
changed date format

Olusola-Opencast:


kkonuganti (approved):

kberec:
@Olusola-Opencast I know the ticket doesn't mention it, but there are some more places the long format is used for the next review date: 
- landing.html
- nextReviewDateQuestion.html
- nextReviewDateStandalone.html
- nextReviewDate.html
Could you please check with Anthony, if there is a separate ticket for these cases?
I think the issue here is that this ticket wasn't 3A'd yet and wasn't in the "Ready to go" Jira bucket, otherwise these questions might have been asked as a part of the 3A.

PR #629, opened 2023-01-17
CAT-1401 update postgres version to 14.3

Olusola-Opencast:
CAT-1401 update postgres version to 14.3

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
     working_directory: ~/app
     docker:
       - image: cimg/node:<< pipeline.parameters.node-version >>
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     working_directory: ~/app
     docker:
       - image: cimg/node:<< pipeline.parameters.node-version >>
-      - image: cimg/postgres:14.0
+      - image: cimg/postgres:14.3
```
kberec, line 64: This doesn't seem to be related with the db refresh job. Does it need to be changed (as it is already 14.0)?

kberec (approved):

PR #633, opened 2023-01-20
Cat 1221 changes to the categories on the cat lite page

Olusola-Opencast:
This is not needed for MVP


kberec:
Superseded with other branch

PR #648, opened 2023-02-09
Cat 1422-Integration Test Added

nnehabharti:
Integration-test added


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/mockapis/Elite2Api.groovy
```diff
@@ -25,41 +25,33 @@ class Elite2Api extends WireMockRule {
...
-            .withHeader('Content-Type', 'application/json')
-            .withBody(JsonOutput.toJson([
-              staffId         : user.staffMember.id,
-              username        : user.username,
-              firstName       : user.staffMember.firstName,
-              lastName        : user.staffMember.lastName,
-              email           : 'itaguser@syscon.net',
-              activeCaseLoadId: caseloadId
-            ]))))
+    this.stubFor(get('/api/users/me')
```
kberec, line 32: Reformatting is generally a good thing, but in this case:

- makes it difficult to see what relevant was exactly changed,
- as we have work going on in multiple branches at the moment, if changes are overlapping, makes conflicts more possible, when merging and updating branches.
We need to check if these are changed in other ongoing branches as well...

nnehabharti: Do you want me amend those then?

kberec: Resolving, as we discussed it offline.

kberec (approved):

PR #655, opened 2023-02-17
add categorydecision page test in women specification

nnehabharti:


integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/model/TestFixture.groovy
```diff
@@ -49,6 +49,7 @@ class TestFixture {
   ]
 
   public static final defaultRatingsU = [
```
kberec, line 51: As the concept is now changed, I suggest renaming this method to `defaultRatingsClosed`

kberec (approved):

PR #665, opened 2023-02-24
Cat 1421 allow ip 194 33 192 3 access to dcs in dev envi

Olusola-Opencast:
Cat 1421 allow ip 194 33 192 3 access to dcs in dev envi

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -23,6 +23,10 @@ generic-service:
     DPS_URL: https://digital-dev.prison.service.justice.gov.uk/
 
   allowlist:
+    ark-nps-hmcts-ttp2: 194.33.192.0/25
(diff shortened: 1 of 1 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -23,6 +23,10 @@ generic-service:
     DPS_URL: https://digital-dev.prison.service.justice.gov.uk/
 
   allowlist:
+    ark-nps-hmcts-ttp2: 194.33.192.0/25
```
kberec, line 26: I'd say to add ark-nps-hmcts-ttp1 too, as it is the same group, even it is a different ip address range.

Olusola-Opencast: added.

kberec (approved):

PR #671, opened 2023-03-02
CAT-1480 Updates to the Women's initial cat stats page

kberec:
Added conditional sections to dashboardInitial view and added specific query for female decisions section


Olusola-Opencast (approved):
ok

PR #672, opened 2023-03-03
Cat 1430 a 11 y link to accessibility statement page

Olusola-Opencast:
CAT-1430 add accessibility statement page and link in footer

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -208,8 +208,9 @@ module.exports = function createApp({
     return res.render('autherror')
   })
 
-  app.get('/accessibilityStatement', (req, res) => {
-    return res.render('accessibilityStatement')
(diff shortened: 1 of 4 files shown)


server/app.js
```diff
@@ -208,6 +208,11 @@ module.exports = function createApp({
     return res.render('autherror')
   })
 
+  app.get('/accessibility-statement', async (req, res) => {
+    const user = await userService.getUser(res.locals)
```
kberec, line 212: Not used in the view. Do we still need to query and pass in the value?

kberec: Yes, user name is displayed via layout.html

kberec (approved):

kberec (approved):

PR #679, opened 2023-03-20
CAT-1504 adds TPRS details to stats pages

a6software:


server/services/statsService.js
```diff
@@ -84,5 +84,22 @@ module.exports = function createstatsService(statsClient) {
         total: onTime + notOnTime,
       }
     },
+
+    async getInitialCategorisationTprsTotals(startDate, endDate, prisonId, transactionalClient) {
```
kberec, line 88: All the other methods have catType as a parameter. Was there a specific reason to have two separate methods here?

server/views/partials/dashboardData.html
```diff
@@ -1,3 +1,13 @@
+  <h3>Temporary Presumptive Recategorisation Scheme (TPRS)</h3>
+  <table class="govuk-table" id="tprs">
+    <tbody class="govuk-table__body">
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">Total {% if catType == 'INITIAL'%}initial{% endif %} categorisations through TPRS</td>
```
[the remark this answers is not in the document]
kberec: Might be a deviation between the screenshots on the ticket and the prototype. I think (but please check with Barry) the prototype is correct.

server/services/statsService.js
a6software: Changed to have a single `getTprsTotals` fn that takes `catType`

kberec (approved):

PR #689, opened 2023-03-27
[CAT-1513] RFC - an approach to testing SQL directly in Jest

a6software:
Draft at the moment to test the pipeline only


a6software:
proof of concept, only needed to validate CircleCI - closing.

PR #690, opened 2023-03-27
CAT-1487 Change condition for higher security review

kberec:


a6software (dismissed):
I don't understand the business case, but the logic seems sound 👍

Olusola-Opencast (commented):
ok

Olusola-Opencast (approved):
ok

PR #691, opened 2023-03-30
[CAT-1518] Adds direct SQL query testing

a6software:
This PR provides an approach to directly testing SQL queries against a test database. 

Comments below discuss various approaches to how that DB is created. One approach has been provided. 

The biggest change here is needing to define the `DB_PORT` which will impact all environments - though a default has been provided. This is because the existing knex setup does not factor in a DB port, and if going with a separate DB we cannot run two instances of Postgres on the same port (`5432` by default).

This may already suggest that a better approach is one DB with an `init` script. This is built into the Docker postgres image, but would mean recreating your DB environment - dump / restore if wanting to keep existing data. 

Typing this up, I'm already swaying towards an init script. 

**Edit** - init script approach PR - https://github.com/ministryofjustice/offender-categorisation/pull/691

### DIFF
--- a/docker-compose-test.yml
+++ b/docker-compose-test.yml
@@ -12,6 +12,21 @@ services:
       - POSTGRES_PASSWORD=form-builder
       - POSTGRES_USER=form-builder
       - POSTGRES_DB=form-builder
+    command: ["postgres", "-c", "log_statement=all", "-c", "log_destination=stderr"]
+
+  form-db-unit-tests:
+    image: postgres
+    networks:
+      - hmpps
+    container_name: form-builder-db-unit-tests
+    restart: always
+    ports:
+      - "5434:5432"
+    environment:
+      - POSTGRES_PASSWORD=form-builder-unit-tests
+      - POSTGRES_USER=form-builder-unit-tests
+      - POSTGRES_DB=form-builder-unit-tests
+    command: ["postgres", "-c", "log_statement=all", "-c", "log_destination=stderr"]
 
   categorisation-redis:
     image: redis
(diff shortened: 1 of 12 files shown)


server/config.js
```diff
@@ -40,6 +40,7 @@ module.exports = {
     password: get('DB_PASS', 'form-builder'),
     server: get('DB_SERVER', 'localhost', true),
     database: get('DB_NAME', 'form-builder', true),
+    port: get('DB_PORT', 5432, true),
```
a6software, line 43: `DB_PORT` is new and the biggest risk here. 

`5432` is the default port for postgres - but do all environments use `5432` ? 

My best guess is this is fine, but I'm not absolutely certain.

docker-compose-test.yml
```diff
@@ -12,6 +12,21 @@ services:
...
+      - hmpps
+    container_name: form-builder-db-unit-tests
+    restart: always
+    ports:
+      - "5434:5432"
+    environment:
+      - POSTGRES_PASSWORD=form-builder-unit-tests
+      - POSTGRES_USER=form-builder-unit-tests
+      - POSTGRES_DB=form-builder-unit-tests
+    command: ["postgres", "-c", "log_statement=all", "-c", "log_destination=stderr"]
```
a6software, line 29: Can either create a new DB for unit testing, or we can manually `create database form-builder-unit-test` inside the existing DB, or we can create an init script for the existing DB to create 2 (or 3, one for risk-profiler) databases on first boot?

a6software:
closing in favour of https://github.com/ministryofjustice/offender-categorisation/pull/692

PR #692, opened 2023-03-31
[CAT-1518] SQL Query unit testing

a6software:
This PR provides an approach to directly testing SQL queries against a test database.

Requires that the docker-compose environment is running in order to run unit tests. This is because of the new reliance on the DB. There is not a huge amount of overhead added, however. All tests still pass locally in approx 6 seconds. 

This approach leverages one DB server for all tests, where that server has multiple databases (created at container initialisation via the init script).


kberec:
Looking good @a6software . Hope I manage to write this comment before access is removed, as we already been signed out of slack, etc, so offboarding has started - might not get a chance to go through in details. Seen Olu highlighted few points on slack. The only thing I'd add is some kind of description to a people who might prefer NOT to run docker on localhost for development. For that scenario they could install the db locally and create this testing db manually.

a6software:
Thanks for your feedback @kberec - much appreciated as always

scripts/init.sql
```diff
@@ -0,0 +1,3 @@
+-- create database "form-builder"; -- this one is automatically created on initial container creation
+create database "form-builder-unit-tests";
+create database "risk-profiler";
```
a6software, line 3: create database "form-builder-integration-tests";

PR #693, opened 2023-04-03
[CAT-1522] Adds Cypress and initial healthcheck test

a6software:

PR #694, opened 2023-04-05
[CAT-1526] Adds login Cypress tests

a6software:
Reliant on https://github.com/ministryofjustice/offender-categorisation/pull/693

This is currently following the setup steps that are in the Groovy tests under the `model` directory. These are now TypeScript factories - however, I suspect this approach will change / adapt as a proper feature spec migration takes place. 

Because of the way cat tool landing page works, this is divergent from the TypeScript starter project for Login, and has already begun to implement part of the Groovy test spec steps.


a6software:
merged via https://github.com/ministryofjustice/offender-categorisation/pull/700

PR #696, opened 2023-04-12
[CAT-1527] Adds security page tests

a6software:


a6software:
merged in https://github.com/ministryofjustice/offender-categorisation/pull/696

PR #697, opened 2023-04-14
[CAT-1528] Adds 'error' page tests

a6software:


a6software:
merged via https://github.com/ministryofjustice/offender-categorisation/pull/700

PR #706, opened 2023-05-05
[CAT-1539] Migrates Cat Decision Groovy test to Cypress

a6software:


a6software:
included in https://github.com/ministryofjustice/offender-categorisation/pull/711

PR #709, opened 2023-05-10
[CAT-1542] Migrate further charges groovy tests to Cypress

a6software:
Note: this adds more tests than in the original groovy file


a6software:
included in https://github.com/ministryofjustice/offender-categorisation/pull/711

PR #722, opened 2023-07-04
[SI-165] Remove womens cats from Cat Lite Cypress test

a6software:
Includes:

* SI-108
* SI-63
* SI-35
* SI-24


server/views/formPages/supervisor/review.html
```diff
@@ -226,14 +229,16 @@ <h2 class="govuk-heading-m">Other Information</h2>
...
           value: "No",
           text: noLabel,
           conditional: {
             html:overriddenCategoryTextHtml
-          }
+          },
+          checked: data.supervisor.review.supervisorCategoryAppropriate === 'No'
         }
       ]
     }) }}
```
a6software, line 244: this is a bug fix - [SI-63]

integration_tests/e2e/ratings/securityInput.cy.ts
```diff
@@ -125,7 +125,7 @@ describe('Security Input', () => {
         })
       })
 
-      describe("it should record a 'yes' decision", () => {
+      describe.only("it should record a 'yes' decision", () => {
```
Stubbs: Is this `.only()` intentional?

Stubbs (approved):
In terms of code itself it looks fine, just check that `.only` on one of the tests, feels like it might be blocking a bunch of tests from running. I don't know the system well enough to comment on what they're covering though, @garethmdavies maybe?

garethmdavies (approved):
LGTM

PR #731, opened 2023-08-22
[SI-264] Rejected recats display when next review is after release

a6software:
Fixes a bug whereby offenders were not showing up on a recat dashboard if their release date was after their next review date, but their recat had been rejected by the supervisor.


test/services/offenderService.test.js
```diff
@@ -3550,3 +3639,204 @@ describe('statusTextDisplay', () => {
...
+      dbRecord: {
+        status: Status.SUPERVISOR_BACK.name,
+      },
+      releaseDate: null,
+      expected: true,
+    },
+  ]
+
+  testCases.forEach(({ description, dbRecord, releaseDate, expected }) => {
+    it(description, () => {
```
a6software, line 3837: appreciate this does not use the `each` approach - this is a copy from the previous PR where these tests were passing and I do not have time to refactor right now

steverendell (approved):
lgtm

a6software:
ty

PR #744, opened 2023-09-25
[SI-288] Migrate to DPS Components for Header & Footer

a6software:


integration_tests/e2e/dpsFrontEndComponents.cy.ts
```diff
@@ -0,0 +1,306 @@
+import DashboardInitialPage from '../pages/dashboard/initial'
+import Page from '../pages/page'
+import { SUPERVISOR_USER } from '../factory/user'
+import { cleanString } from '../support/utilities'
+
+describe.skip('DPS Front End Components', () => {
```
a6software, line 6: deliberately skipped as this will not work until we migrate

covered in https://dsdmoj.atlassian.net/browse/SI-409

cruffymoj (approved):
It is to verbose and abstract to understand this PR, ive gone over it a few times, think its best to getit on dev and demo and test

a6software:
cheers

PR #746, opened 2023-09-26
[BUG] Fixup invalid number of arguments in env get function

a6software:


Stubbs (approved):
LGTM.

I'd probably put a default of false on log as well, just to reduce the risk leaking anything into the logs, if you want to log you then have to be very specific to do so.

a6software:
> LGTM.
> 
> I'd probably put a default of false on log as well, just to reduce the risk leaking anything into the logs, if you want to log you then have to be very specific to do so.

hmm I solved this in a PR that should have come later - https://github.com/ministryofjustice/offender-categorisation/pull/747

PR #748, opened 2023-09-26
[SI-288] Adds feature flag middleware and display helper

a6software:


Stubbs (approved):
LGTM

I can see how we can easily extend it in the future to allow easier configuration of feature flags with the config/middleware, maybe even a db table to allow a GUI to control which features are on/off.

PR #754, opened 2023-10-05
[SI-288] Removes feature flag wrapper

a6software:


server/app.js
```diff
@@ -21,7 +20,6 @@ const createFormRouter = require('./routes/form')
 const createTasklistRouter = require('./routes/tasklist')
 const createTasklistRecatRouter = require('./routes/tasklistRecat')
 const authorisationMiddleware = require('./middleware/authorisationMiddleware')
-const featureFlagMiddleware = require('./middleware/featureFlagMiddleware')
```
Stubbs, line 24: Is this not required any more? How do the feature flags work?

a6software: It was added only to enable feature flagging this specific change.

This change removes the feature flagging as the change is now live and 'on' permanently. 

There is nothing else using the FF stuff at the time, so for me it's dead code.

PR #755, opened 2023-10-05
[SI-427] Back to case list should take the user to the site root

a6software:


integration_tests/fixtures/liteCategoriser/supervisorView.ts
```diff
@@ -15,7 +15,7 @@ export const supervisorViewSeedData: LiteCategoryDbRow[] = [
     approved_by: null,
     assessment_committee: 'RECP',
     assessment_comment: 'comment',
-    next_review_date: moment().add(6, 'months').startOf('day').toISOString(),
+    next_review_date: moment().add(6, 'months').toISOString(),
```
a6software, line 18: odd one, this is what flagged the test as failing, but when fixed the other issue started below

server/views/formPages/approvedView.html
```diff
@@ -156,7 +156,7 @@ <h2 class="govuk-heading-m">Other information</h2>
   </div>
 
 {{ govukButton({
-  href: backLink,
+  href: "/",
```
a6software, line 159: `backLink` is not passed in to the template. I can't understand how this was ever a passing test.

PR #760, opened 2023-10-18
SI-399-hide-PII-titles-from-google-analytics

jonhindle:
Add override to GA titles to prevent them from recording prisoner details


server/views/partials/layout.html
```diff
@@ -12,7 +12,7 @@
       window.dataLayer = window.dataLayer || [];
       function gtag(){dataLayer.push(arguments);}
       gtag('js', new Date());
-      gtag('config', '{{ googleAnalyticsKey }}');
+      gtag('config', '{{ googleAnalyticsKey }}', {{ { title: '' } if googleAnalyticsOverrideTitle else {} }});
```
Stubbs, line 15: Will this switch all analytics off for when override is true?

server/views/formPages/approvedView.html
```diff
@@ -1,6 +1,7 @@
 {% extends "./formTemplate.html" %}
 {% from "summary-list/macro.njk" import govukSummaryList %}
 
+{% set googleAnalyticsOverrideTitle = true %}
```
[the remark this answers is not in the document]
jonhindle: Can do 🤷 
Blanking the title for now seems cleanest; if required we can add in override values later when we've got a clearer idea of what we want to do with the analytics.

server/views/partials/layout.html
jonhindle: No, it should just override the title value sent to GA.

Stubbs (approved):
LGTM

PR #764, opened 2023-11-07
fix-feComponents-typo

jonhindle:
Fix the typo in `layout.html` that should say `if feComponents.jsIncludes` with an 's' on the end. JS currently not working in the app


a6software:
> JS currently not working in the app

This means JS coming via front end components API isn't working?

I tested all the site specific JS that I am aware of and it seems working as expected. 

Sorry about that all the same!

a6software (approved):

whitfield-mj:
@a6software the error was in the components library documentation for a while, not your fault

PR #778, opened 2024-01-18
SDIT-1410 UserResource use client token

steverendell:
A prison-api endpoint needs to switch form auth token to client token

### DIFF
--- a/docker-compose-test.yml
+++ b/docker-compose-test.yml
@@ -73,7 +73,7 @@ services:
     image: wiremock/wiremock
     networks:
       - hmpps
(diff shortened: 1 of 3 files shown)


docker-compose-test.yml
```diff
@@ -73,7 +73,7 @@ services:
     image: wiremock/wiremock
     networks:
       - hmpps
-    container_name: wiremock
+    container_name: categorisation-wiremock
```
steverendell, line 76: avoid clash with other projects

a6software (approved):

PR #781, opened 2024-01-26
[SI-607] Log out prisoner booking IDs assumed needing removing from lite_category table

a6software:
The aim of this PR is not to actually remove the prisoners, but to list those who _would_ be removed, once the results have been manually verified.


Stubbs (approved):
LGTM

PR #783, opened 2024-01-31
Bug/si 607 remove released prisoners from cat lite approve list

a6software:


Stubbs (approved):
Approved, but surely there's a better way in the future ... lets get the devs together later and see what we can come up with?

cruffymoj (approved):

a6software:
closed in favour of console log to container

PR #785, opened 2024-02-01
[SI-607] filters prisoners in the release date map, and meet the release date criteria

a6software:


server/services/offendersService.js
```diff
@@ -620,25 +621,16 @@ module.exports = function createOffendersService(
...
-        )
-        .map(p => p.bookingId)
-
-      if (alreadyReleased.length) {
-        logger.debug(
-          'The following prisoners have been released and should be removed from the lite_category table',
-          alreadyReleased
-        )
+      if (released.length) {
+        logger.debug('The following prisoners should be removed from the lite_category table', released)
```
jonhindle, line 629: Minor comment, do we lose any context with the shorter debug message?
> "...have been released and..."

jonhindle (approved):
lgtm

server/services/offendersService.js
a6software: The idea here is that the log is short lived - just long enough to give us confidence that running a `delete` query against those `bookingId` values from the `lite_category` table will remove the rows we expect, and no more. 

From a certain point of view, you could say the message now is slightly more correct, in that I found a prisoner who had died in prison, so idk if they were ever released but they definitely got removed from the `releaseDate` map.

This message never actually made it further than pre-prod either, so no change to what is logged in prod.

PR #791, opened 2024-02-09
[SI-732] Fixes conditional display inputs using radio buttons

a6software:


Stubbs (approved):
LGTM - Removes extra CSS class after GDS lib upgrade.

PR #797, opened 2024-03-27
Chore/si 701 gds upgrade

timothy-ness:
GDS upgrade from 4.8 -> 5.2


server/views/pages/categoriserHome.html
```diff
@@ -76,7 +76,7 @@ <h1 class="govuk-heading-l">Prisoner Categorisation</h1>
       <td class="govuk-table__cell">
         {% if row.displayStatus == 'Back from Supervisor' %}
           {{ govukTag({
-            html: 'Rejected by<br/>supervisor'
+            html: 'REJECTED BY<br/>SUPERVISOR'
```
Stubbs, line 79: Is this supposed to have been capitalised?

Stubbs (approved):
LGTM

server/views/pages/categoriserHome.html
timothy-ness: Yes, it looks (from the tests) that there must have used to be css which capitalised this, and that must have been removed with this upgrade, since the test expects it to be in all caps I've capitalised it

Stubbs: 👍

PR #801, opened 2024-04-08
SI 151 store approved category comment in db

timothy-ness:
It was noticed that approvedCategoryComment wasn't being stored in the database, this PR adds a column to the lite category table to store this field alongside the other form data from that submit.


Stubbs (approved):
LGTM

PR #812, opened 2024-05-28
Bug/si 942 moic 404s

timothy-ness:
To reduce (but not eliminate) the number of 404s from MOIC (also called allocation manager) by only querying the offenders which are actually going to be shown on the page (i.e. post all the filtering).

This will mean the moic calls will no longer be done in parallel with 2 other calls (prisoner search and get security referrals), however it will also reduce the number of calls being made to moic (it does them in batches of 20 at a time) so it should be a net reduction in time. On dev, it reduced the calls from 692 -> 253 for one page load, and reduced the 404s from 297 to 64.


a6software (approved):
👍

Stubbs (approved):

PR #825, opened 2024-07-25
[SI-1091] Delete pending cats on Offender Release event

a6software:
Listen for Offender Release events. Delete any pending (not APPROVED or CANCELLED) cats, recats, and lite cats for the offender.

This is currently behind an inactive feature flag, and for the moment will only log what we would have deleted.

### DIFF
--- a/server/services/sqsService.js
+++ b/server/services/sqsService.js
@@ -115,7 +115,7 @@ module.exports = function createSqsService(offenderService, formService) {
             }
             const reason = event?.additionalInformation?.reason
             if (!reason || reason !== 'RELEASED') {
-              logger.warn({ MessageId: message.MessageId, nomsNumber }, 'Reason was not "RELEASED"')
+              logger.info({ MessageId: message.MessageId, nomsNumber }, 'Reason was not "RELEASED"')
               break
             }
             await formService.deletePendingCategorisations(nomsNumber, transactionalDbClient)
--- a/server/services/sqsService.js
+++ b/server/services/sqsService.js
@@ -4,6 +4,7 @@ const logger = require('../../log')
 const config = require('../config')
 const riskChangeHelper = require('../utils/riskChange')
 const db = require('../data/dataAccess/db')
+const { events } = require('../utils/eventUtils')
 
 AWS.config.update({
   region: 'eu-west-2',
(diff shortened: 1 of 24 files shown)


server/services/sqsService.js
```diff
@@ -102,9 +103,19 @@ module.exports = function createSqsService(offenderService, formService) {
...
+            logger.info(
+              { event },
+              `${events.EVENT_DOMAIN_PRISONER_OFFENDER_SEARCH_PRISONER_RELEASED}: Received payload`
+            )
+            const nomsNumber = event?.additionalInformation?.nomsNumber
+            if (!nomsNumber) {
+              logger.warn({ MessageId: message.MessageId }, 'MessageId was missing a nomsNumber')
+              break
+            }
+            await formService.deletePendingCategorisations(nomsNumber, transactionalDbClient)
```
a6software, line 116: INCORRECT - only do this if 'reason' is RELEASED

a6software: fixed

server/services/sqsService.js
```diff
@@ -102,9 +103,24 @@ module.exports = function createSqsService(offenderService, formService) {
...
+              `${events.EVENT_DOMAIN_PRISONER_OFFENDER_SEARCH_PRISONER_RELEASED}: Received payload`
+            )
+            const nomsNumber = event?.additionalInformation?.nomsNumber
+            if (!nomsNumber) {
+              logger.warn({ MessageId: message.MessageId }, 'MessageId was missing a nomsNumber')
+              break
+            }
+            const reason = event?.additionalInformation?.reason
+            if (!reason || reason !== 'RELEASED') {
+              logger.warn({ MessageId: message.MessageId, nomsNumber }, 'Reason was not "RELEASED"')
```
timothy-ness, line 118: Should this be a warning? Is it something which could happen often or a kind of error?

timothy-ness (dismissed):

server/services/sqsService.js
a6software: good spot, thanks - no, lazy copy paste from the nomsNumber check above, fixed (set to info) and pushed

timothy-ness (approved):

PR #830, opened 2024-08-29
Recategorisation prioritisation filter frontend

timothy-ness:
This is the frontend part of the recategoriserHome page's new prioritisation filter. It isn't actually applying the filter but you can interact with it fully, hiding the filter, selecting all, or individually selecting the inputs. The 'backend' part of actually applying the filters will be in a separate PR.


server/views/pages/recategoriserHome.html
```diff
@@ -42,81 +45,196 @@ <h1 class="govuk-heading-l">{{ pageTitle }}</h1>
...
+                text: 'Clear filters',
+                href: '/recategoriserHome'
+              },
+              categories: filterCategories
+            } if filterCategories.length > 0,
+            optionsHtml: filterOptionsHtml,
+            submit: {
+              attributes: {
+                "data-test-id": "submit-button",
+                style: "display: none;"
```
a6software, line 135: Screenshot 2024-09-02 at 10 17 07

i tried it without line 135 and it seems hidden by default anyway, so maybe you can just remove this line. If not, it's probably easiest to give this element an id, and then add a style to hide it from the stylesheet, which is valid for the csp

assets/js/recategoriserHome.js
```diff
@@ -0,0 +1,79 @@
...
+  const csrfToken = document.querySelector('meta[name="csrf-token"]').getAttribute('content')
+  await fetch('/recategoriserHome/hide-filter', {
+    method: 'POST',
+    body: JSON.stringify({ hideFilter }),
+    headers: {
+      'Content-Type': 'application/json',
+      'X-CSRF-Token': csrfToken,
+    },
+  });
+})
```
a6software, line 79: it's not going to make much difference here I imagine, but you can (often should) store the lookup result in a variable, like

```
const tableContainer = document.getElementById('tableContainer')
const filterContainer = document.getElementById('filterContainer')

  if (event.target.innerText === 'Hide filter') {
    filterContainer.classList.add('govuk-visually-hidden')
    tableContainer.classList.remove("govuk-grid-column-two-thirds")
    tableContainer.classList.add("govuk-grid-column-full")
```

as otherwise it does the same lookup multiple times.

assets/js/recategoriserHome.js
```diff
@@ -0,0 +1,79 @@
...
+  return true
+}
+
+function setSelectAllButtonTextBasedOnSelectedCheckboxes(selectAllButtonId, checkboxesIdPrefix) {
+  document.getElementById(selectAllButtonId).innerText = allCheckboxesAreChecked(checkboxesIdPrefix) ? 'Deselect all' : 'Select all'
+}
+
+document
+  .getElementById('selectAllSuitabilityForOpenConditionsFilter')
+  .addEventListener(
```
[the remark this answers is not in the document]
a6software: this would be better being:

```
const selectAllSuitabilityForOpenConditionsFilter = document
  .getElementById('selectAllSuitabilityForOpenConditionsFilter')

if (selectAllSuitabilityForOpenConditionsFilter) {
  selectAllSuitabilityForOpenConditionsFilter.addEventListener(...
```

a6software (approved):
lgtm

PR #831, opened 2024-08-29
[SI-607] Adds cypress test for Categoriser Home page

a6software:
Intended to be a Cypress replica of https://github.com/ministryofjustice/offender-categorisation/blob/main/integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/HomePageSpecification.groovy#L20-L69


integration_tests/mockApis/elite2.ts
```diff
@@ -645,6 +647,34 @@ const stubGetOffenderDetails = ({
   return Promise.all([stubBasicInfo(), stubSentenceDetail(), stubSentenceTerms(), stubMainOffence()])
 }
 
+const stubGetOffenderDetailsBasic = ({
```
timothy-ness, line 650: Could you call stubGetOffenderDetails here with basic parameter set to true? The json body here overlaps with that functions, if basic is true should it be this smaller payload?

timothy-ness (dismissed):

integration_tests/mockApis/elite2.ts
a6software: implemented as suggested

integration_tests/mockApis/allocationManager.ts
```diff
@@ -21,7 +21,7 @@ export default {
     stubFor({
       request: {
         method: 'GET',
-        urlPattern: '/api/allocation/\\w+',
+        urlPattern: '/allocation-manager/api/allocation/\\w+',
```
a6software, line 24: the path was wrong here, which is why Engelbert Humperdinck now needed adding anywhere POM is displayed - it should always have displayed, but the mock was not properly setup.

timothy-ness (approved):

a6software:
thanks

PR #833, opened 2024-09-02
[SI-607] Filter prisoners from cat / recat list by legal status

a6software:
Intentionally duplicates cat and recat home pages, though tweaks the lists to filter out any unwanted offenders based on their legal status.

This is duplicated such that we can display the changes to a small / controlled subset of users, before rolling out to the wider estate. At that point, the duplicated code can be removed.


a6software:
changes requested

timothy-ness (approved):

a6software:
thanks

PR #837, opened 2024-09-09
Recategorisation prioritisation backend

timothy-ness:
Applying the filters when they are selected, except the RoSH filter because that is still waiting on credentials being added to our client ID. If there are no results and filters are applied then it will show a different message compared to normal no results. 

Ticket: https://dsdmoj.atlassian.net/browse/SI-1078


server/services/offendersService.js
```diff
@@ -742,12 +762,15 @@ module.exports = function createOffendersService(
           return null
         }
 
-        const releaseDate = releaseDateMap.get(raw.bookingId)
+        const releaseDateForDecidingIfRecordShouldBeIncluded =
+          !raw.dbRecord || !raw.dbRecord.catType || raw.dbRecord.catType === CatType.RECAT.name
```
timothy-ness, line 766: This is to match the logic on line 57, since a different function is now being called to get the release date (getprisonerSearchData).

server/services/recategorisationFilter.ts
```diff
@@ -0,0 +1,161 @@
...
+  const prisonerNumbers = prisoners.map(prisoner => prisoner.offenderNo)
+  const dateThreeMonthsAgo = getDateNMonthsAgo(3)
+  const dateTwoMonthsAgo = getDateNMonthsAgo(2)
+  const dateOneMonthAgo = getDateNMonthsAgo(1)
+  const [adjudicationsThreeMonthsAgo, adjudicationsTwoMonthsAgo, adjudicationsLastMonth] = await Promise.all([
+    nomisClient.getOffenderAdjudications(prisonerNumbers, dateThreeMonthsAgo, dateTwoMonthsAgo, agencyId),
+    nomisClient.getOffenderAdjudications(prisonerNumbers, dateTwoMonthsAgo, dateOneMonthAgo, agencyId),
+    nomisClient.getOffenderAdjudications(prisonerNumbers, dateOneMonthAgo, moment().format('YYYY-MM-DD'), agencyId),
+  ])
+  return [...adjudicationsThreeMonthsAgo, ...adjudicationsTwoMonthsAgo, ...adjudicationsLastMonth]
```
a6software, line 76: i don't follow this - so may very well have missed something

is this possibly simplified with a single call from 

` nomisClient.getOffenderAdjudications(prisonerNumbers, dateThreeMonthsAgo, moment().format('YYYY-MM-DD'), agencyId),`

server/services/recategorisationFilter.ts
```diff
@@ -0,0 +1,161 @@
...
+  return prisoners.filter(prisoner => {
+    const currentPrisonerSearchData = prisonerSearchData.get(prisoner.bookingId)
+    const activeNonExpiredAlertCodes =
+      (currentPrisonerSearchData &&
+        currentPrisonerSearchData.alerts
+          ?.filter(alert => alert.active && !alert.expired)
+          .map(alert => alert.alertCode)) ||
+      []
+    const incentiveLevelCode = currentPrisonerSearchData?.currentIncentive?.level.code
+    for (let i = 0; i < allFilters.length; i += 1) {
```
a6software, line 104: possibly slightly nicer to use `for (const filter of allFilters)`

a6software (dismissed):

server/services/recategorisationFilter.ts
timothy-ness: I get an eslint error if I use for of: iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them

a6software (approved):

PR #844, opened 2024-09-17
[SI-1206] Adds xlsx for spreadsheet output, db cert path update for same

a6software:
Pre-requisite change, to enable testing of scripts on pre-prod before committing the full code.

### DIFF
--- a/server/data/dataAccess/db.js
+++ b/server/data/dataAccess/db.js
@@ -14,7 +14,7 @@ const pool = new Pool({
   ssl:
     config.db.sslEnabled === 'true'
       ? {
-          ca: fs.readFileSync('root.cert'),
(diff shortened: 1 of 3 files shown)


server/data/dataAccess/db.js
```diff
@@ -14,7 +14,7 @@ const pool = new Pool({
   ssl:
     config.db.sslEnabled === 'true'
       ? {
-          ca: fs.readFileSync('root.cert'),
+          ca: fs.readFileSync('/app/root.cert'),
```
a6software, line 17: so i can use the same db config from a new `scripts` dir outside the conventional `server` dir

timothy-ness (approved):

a6software:
thx

PR #845, opened 2024-09-17
[SI-1206] Extract and export TPRS stats

a6software:


scripts/tprs-data-extract/1211-three-years-or-more.js
```diff
@@ -0,0 +1,116 @@
...
+    const bookings = await getCatCPrisoners({ onlyTprs, startDate, endDate })
+
+    const bookingMap = createBookingsMap({ bookings })
+    const bookingIds = bookingMap.keys()
+
+    const prisonerSearchData = await getPrisonerSearchData({ username, bookingIds })
+    const updatedBookingMap = updateBookingMapWithSearchResult({ bookingMap, prisonerSearchData })
+    const threeYearsFromNow = moment().add(3, 'years')
+
+    return [...updatedBookingMap.values()]
```
timothy-ness, line 94: Do you need to splay this into an array? does updatedBookingMap.values().filter not work?

timothy-ness (approved):

cruffymoj (changes requested):
Added @garethmdavies

garethmdavies (changes requested):
For this type of requirement, the analysis should really be done on the analytical platform and not within the operational system. As such we should support getting the cat tool data to the AP via the DPR platform for the data analysts to provide this to DOS.

Where MI / metrics are required in service, these should now be done using DPR embedded reporting components and not standalone SQL scripts which will have an additional support and maintenance burden for the team.

See: https://dsdmoj.atlassian.net/wiki/spaces/DPR/pages/4396712508/Embeddable+Reporting

@atulloch

PR #848, opened 2024-10-07
Add azure analytics

timothy-ness:
Adding web analytics to azure for the cat tool. After a talk given at one of the show and tells, some of the team were interested in replicating the dashboards another service have created on Azure for our apps, this is the first step, recording the data in azure.


a6software (approved):
nice, looking forward to seeing what this produces

PR #854, opened 2024-10-14
Si 1258 async middleware non transaction

timothy-ness:
Creating an async middleware which is not in a DB transaction and switching some of the controller actions to use that.


a6software (approved):
ok - so I think this is right. I've tried locally, and I've run all the tests. 

I don't know a definitive way to hammer it to test. 

My opinion is this goes into a release. Just this. Clear the pipelines, and deploy only this one. And let it sit in prod for at least a couple of days before we start pushing out more prod stuff.

a6software (approved):

PR #858, opened 2024-10-18
Handling restricted access people

timothy-ness:
It seems that we get a 403 for a restricted access person, when that happens we should continue as if they have no rosh but record it so we can see how often it happens.


server/data/risksAndNeeds/risksAndNeedsApi.ts
```diff
@@ -70,6 +70,11 @@ function risksAndNeedsApiGetBuilder(username) {
       if (error.response?.status === 404) {
         return {}
       }
+      // 403 errors are returned for crns of restricted access people, we will record how many these occur for but continue as if they do not have a rosh score
+      if (error.response?.status === 403) {
+        logger.info('Risks and needs API: restricted crn requested')
```
a6software, line 75: worth logging the crn here, or do you already have it?

a6software (approved):

server/data/risksAndNeeds/risksAndNeedsApi.ts
timothy-ness: It's in the request since it's a get request

PR #860, opened 2024-10-21
Added in logging and slight refactor to improve performance

timothy-ness:
Improvements for performance:
- ordering the filters so that it will apply all filters which don't require more data to be loaded first, and using the filtered list of prisoners to load data for the further filters.
- increased the batch size of the rosh data so more will be done in parallel.

logging added:
- to see if it's using the cache for RoSH correctly
- to see how long it takes to complete the fetching of crns and full fetching of RoSH data
- to see how many crns have no RoSH level

### DIFF
--- a/server/services/recategorisation/filter/recategorisationFilter.ts
+++ b/server/services/recategorisation/filter/recategorisationFilter.ts
@@ -244,7 +244,7 @@ export const filterListOfPrisoners = async (
 
   let offenderNumbersWithAdjudications = []
   if (allFilters.includes(NO_ADJUDICATIONS_IN_THE_LAST_3_MONTHS)) {
-    const adjudicationsData = await loadAdjudicationsData(prisoners, nomisClient, agencyId)
+    const adjudicationsData = await loadAdjudicationsData(filteredPrisoners, nomisClient, agencyId)
     offenderNumbersWithAdjudications = adjudicationsData.map(adjudicationsDatum => adjudicationsDatum.offenderNo)
     filteredPrisoners = filteredPrisoners.filter(prisoner => {
       return !offenderNumbersWithAdjudications.includes(prisoner.offenderNo)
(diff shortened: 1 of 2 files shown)


server/services/recategorisation/filter/recategorisationFilter.ts
```diff
@@ -244,9 +236,31 @@ export const filterListOfPrisoners = async (
...
+  if (allFilters.includes(NO_ADJUDICATIONS_IN_THE_LAST_3_MONTHS)) {
+    const adjudicationsData = await loadAdjudicationsData(filteredPrisoners, nomisClient, agencyId)
+    offenderNumbersWithAdjudications = adjudicationsData.map(adjudicationsDatum => adjudicationsDatum.offenderNo)
+    filteredPrisoners = filteredPrisoners.filter(prisoner => {
+      return !offenderNumbersWithAdjudications.includes(prisoner.offenderNo)
+    })
+  }
+  let offenderNumbersWithLowRoshScore = []
+  if (allFilters.includes(LOW_ROSH)) {
+    offenderNumbersWithLowRoshScore = await getOffenderNumbersWithLowRoshScore(
```
a6software, line 255: is this or line 247 slow?

if so, it could be worth just adding the functions to an array

```
  if (allFilters.includes(NO_ADJUDICATIONS_IN_THE_LAST_3_MONTHS)) {
    const adjudicationsDataPromise = loadAdjudicationsData(filteredPrisoners, nomisClient, agencyId)
  }
  if (allFilters.includes(LOW_ROSH)) {
    offenderNumbersWithLowRoshScorePromise =getOffenderNumbersWithLowRoshScore(...
}
```

then, run all the promises at the same time with 

```
const [adjudicationsData, offenderNumbersWithLowRoshScore] = await Promise.all([adjudicationsDataPromise, offenderNumbersWithLowRoshScorePromise])
```

requires jiggery pokery with the code after

but then everything runs rather than blocks and then runs and then blocks and then runs etc

a6software (approved):

server/services/recategorisation/filter/recategorisationFilter.ts
timothy-ness: Thanks, good suggestion! I think the adjudications data is normally quite quick to load but this is still a good call... I'm going to add it in as part of a different PR though so I can run some of this logging on preprod sooner

PR #866, opened 2024-10-29
SI 1290 prison ID feature flag and logging for analytics

timothy-ness:
Allowing us to enable the feature by prison ID so we can switch it on for the pilot prisons. Also adding in logging to be able to show report on a couple of the analytics questions UCD want to be able to answer:
- Do users remove the filter chips? If so, which ones?
- What is the average number of cases revealed, after all filters are applied? For a prison and for a single pom?
- How frequently do no results occur?


server/utils/nunjucks.utility.ts
```diff
@@ -14,19 +14,24 @@ export const removeFilterFromFullUrl = (
...
-      fullUrl.length
-    )
-  )
+    url.substring(
+      url.indexOf(filter) + filter.length + (url[startPositionOfFilterInUrl - 1] === '&' ? 0 : 1),
+      url.length
+    ) +
+    (numberOfFiltersApplied > 1 ? '&' : '?')
+  }filterRemoved=${filter}`
 }
```
a6software, line 35: could this be done using `URL`? 

```
const url = new URL(fullUrl);

if (url.searchParams.has('filterRemoved')) {
  url.searchParams.delete('filterRemoved')
}

// ... and so on
```

a6software (dismissed):

cruffymoj (approved):

a6software (approved):

PR #884, opened 2024-12-17
[SI-1317] Migrate openConditions groovy test to cypress - first test only

a6software:
There is no behaviour change here - yet.

This test aims to replicate all of this - https://github.com/ministryofjustice/offender-categorisation/blob/main/integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/OpenConditionsSpecification.groovy#L41-L371

however, I have deliberately removed the two 'in test sub tests' where validation failures are tested. Those are to be added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,9 +79,9 @@
     "delay": 2500
   },
   "dependencies": {
+    "@joi/date": "^2.1.1",
     "@microsoft/applicationinsights-clickanalytics-js": "^3.1.0",
     "@microsoft/applicationinsights-web": "^3.1.0",
-    "@joi/date": "^2.1.1",
     "@ministryofjustice/frontend": "^2.1.0",
     "@snyk/protect": "^1.1293.0",
     "agentkeepalive": "^4.2.1",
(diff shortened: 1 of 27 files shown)


tsconfig.json
```diff
@@ -13,11 +13,7 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "experimentalDecorators": true,
-    "typeRoots": [
-      "./server/@types",
-      "./node_modules/@types"
-    ]
```
a6software, line 20: these dirs cause me all manner of chaos locally - server types doesnt exist anyway

package.json
```diff
@@ -79,9 +79,9 @@
     "delay": 2500
   },
   "dependencies": {
+    "@joi/date": "^2.1.1",
     "@microsoft/applicationinsights-clickanalytics-js": "^3.1.0",
     "@microsoft/applicationinsights-web": "^3.1.0",
-    "@joi/date": "^2.1.1",
```
a6software, line 84: think this is prettier's doing

timothy-ness (approved):

PR #886, opened 2025-01-07
Si 1378 security review api

timothy-ness:
Calling an endpoint on hmpps-offender-categorisation-api to insert the security review into the DB rather than inserting it directly within this app, plus all the test refactors that go along with this.


integration_tests/mockApis/elite2.ts
```diff
@@ -1506,6 +1506,27 @@ const stubRecategorise = (
   return Promise.all([recategorisationsStub(agencyId), latestOnlyStub()])
 }
 
+const latestOnlyForOffenderStub = ({ offenderNumber }: { offenderNumber: string }) =>
+  stubFor({
+    request: {
+      method: 'GET',
+      url: `/elite2/api/offender-assessments/CATEGORY?offenderNo=${offenderNumber}&latestOnly=false&activeOnly=false`,
```
a6software, line 1513: is this expected - latestOnly=false - based on the fn name it reads as if that would be true?

integration_tests/db/queries.ts
```diff
@@ -301,6 +310,20 @@ async function updateRiskProfile({
   return await db.query(`update form set risk_profile = $1::JSON where booking_id = $2`, [riskProfile, bookingId])
 }
 
+const updateFormRecord = async ({ bookingId, status, formResponse }: { bookingId: number, status: string, formResponse: any }) => {
+  const existingFormResponse = await db.query(`select form_response from form where booking_id = $1`, [bookingId])
+  return await db.query(
+    `update form set status = $1, form_response = $2 where booking_id = $3`,
+    [status, JSON.stringify({ ...existingFormResponse.rows[0]?.form_response, ...formResponse }), bookingId]
```
a6software, line 317: probably not a concern here - not sure why ramda mergeRight is imported above, but could maybe also be used here?

the outcome of merging two json objects is not a deep merge, but a replacement at the key level. Ramda mergeRight is probably a deep merge - the intuitive outcome. From what I can see of usage here below, it's fine as is.

a6software (commented):
If you're happy with pg-format and the other things, I will approve

PR #887, opened 2025-01-07
[SI-1317] Migrate openConditions groovy test to cypress - second test only

a6software:
It's large because what it migrates from is equally very large - https://github.com/ministryofjustice/offender-categorisation/blob/main/integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/recat/OpenConditionsSpecification.groovy#L203-L362


RAGQYFTJ (approved):

timothy-ness (approved):

a6software:
thanks

PR #905, opened 2025-02-24
Si 1471 add additional logging for error due cats

timothy-ness:
We know that some are errors due to being on remand and some are errors due to being recalls but I want to see if the recall flag or the legal status will help us differentiate between the two so we can handle them correctly.


espedzhev (approved):
lgtm

RAGQYFTJ (approved):

PR #908, opened 2025-02-26
SI-1495 - Overdue text

RAGQYFTJ:


server/services/offendersService.js
```diff
@@ -1117,9 +1119,25 @@ module.exports = function createOffendersService(
...
+    // eslint-disable-next-line default-case
+    switch (true) {
+      case diffInDays > 1:
+        overdueText = `${diffInDays} days overdue`
+        break
+      case diffInDays === 1:
+        overdueText = '1 day overdue'
+        break
+    }
+    return overdueText
```
timothy-ness, line 1140:
```suggestion
    return `${diffInDays} day${diffInDays > 1 ? 's' : ''} overdue`
```

server/views/pages/recategoriserHome.html
```diff
@@ -79,7 +79,7 @@ <h1 class="govuk-heading-l">{{ pageTitle }}</h1>
             <td class="govuk-table__cell">
               {% if row.overdue %}
                 {{ govukTag({
-                  text: 'Overdue',
+                  text: row.overdueText +' overdue',
```
timothy-ness, line 82: Why does this one need the addition of ' overdue' when the other one doesn't?

server/services/offendersService.js
RAGQYFTJ: diffInDays can sometimes be negative. I will have a think about this how to write in a concise manner

server/views/pages/recategoriserHome.html
RAGQYFTJ: good spot, I think I missed removing this in my last push as I was making changes. Fixed

server/services/offendersService.js
RAGQYFTJ: I did this but had to suppress eslint as its nested ternary operator
    // eslint-disable-next-line no-nested-ternary
    return diffInDays > 1 ? `${diffInDays} days overdue` : diffInDays === 1 ? '1 day overdue' : ''

timothy-ness (approved):

PR #917, opened 2025-03-10
Added JSDoc annotations and fix incorrect option key in server config.js

espedzhev:


server/config.js
```diff
@@ -208,5 +221,5 @@ module.exports = {
       },
     },
   },
-  appInsightsConnectionString: get('APPLICATIONINSIGHTS_CONNECTION_STRING', '', { requiredInProduction: true }),
+  appInsightsConnectionString: get('APPLICATIONINSIGHTS_CONNECTION_STRING', '', true, { requireInProduction: true }),
```
espedzhev, line 224: @timharrison-moj Not sure if this should be logging out or not. It was before because the passed object was evaluated as truthy. If it shouldn't, we can quickly change it.

timothy-ness (approved):

PR #919, opened 2025-03-11
CAT Tool - review date pages updates for 3 to 5 years policy changes

espedzhev:
[CAT Tool - review date pages updates for 3 to 5 years policy changes](https://dsdmoj.atlassian.net/browse/SI-1449)


espedzhev:
rebuild

server/views/formPages/nextReviewDate/nextReviewDateQuestion.html
```diff
@@ -18,48 +19,55 @@
...
-    html: guidance
-  }) }}
+  {% if featurePolicyChangeThreeToFiveEnabled %}
+    {{ threeToFivePolicyChangeAlert() }}
+  {% endif %}
+
+  <h1 class="govuk-heading-l govuk-radios--inline">{{ pageTitle }}</h1>
+  <p>Their conditional release date is {{ getLongDateFormat(dateConverter(data.details.sentence.conditionalReleaseDate)) if data.details.sentence.conditionalReleaseDate else '[unavailable]' }}.</p>
+
+    {% set isWithinFiveYears = featurePolicyChangeThreeToFiveEnabled and data.details.sentence.sentenceExpiryDate | isWithinFiveYears %}
```
timothy-ness, line 29: Maybe you were told otherwise by Neena but I thought this was to be the conditionalReleaseDate which is being displayed above?

server/utils/fieldValidation.js
```diff
@@ -75,11 +77,17 @@ function createSchemaFromConfig(pageConfig) {
     indeterminateCheck: (requiredItem = 'indeterminate', requiredAnswer = 'true') =>
       joi.when(requiredItem, {
         is: requiredAnswer,
-        then: joi.date().format('D/M/YYYY').min(today).max(threeYears).required().messages({
-          'date.format': 'The review date must be a real date',
-          'date.max': 'The date that they are reviewed by must be within 3 years',
-          'date.min': 'The review date must be today or in the future',
-        }),
+        then: joi
```
timothy-ness, line 80: I don't think this change is needed, even after this policy change has taken place the indeterminate sentence prisoners should still have a review within 3 years, not 5 years. I'm pretty sure it's just the review frequency which is changing from 1 year to 6 months for prisoners with between 5 and 3 years left to serve.

server/views/formPages/nextReviewDate/nextReviewDateStandalone.html
```diff
@@ -16,19 +17,39 @@
...
-  <h1 class="govuk-heading-l">Change the review date</h1>
-  {% if data.details.sentence.conditionalReleaseDate and not data.details.sentence.indeterminate %}
-    <p class="govuk-!-font-weight-bold govuk-!-margin-bottom-0">Their conditional release date is:</p>
-    <p>{{ getLongDateFormat(dateConverter(data.details.sentence.conditionalReleaseDate)) }}</p>
-  {% elif data.details.sentence.indeterminate %}
-    <p>They have an indeterminate prison sentence.</p>
-  {% endif %}
-  <p>You can alter the prisoner's next review date (Re-assessment date) without needing to create a new categorisation.</p>
+  {% if featurePolicyChangeThreeToFiveEnabled %}
+    <h1 class="govuk-heading-l">Change the review date</h1>
```
timothy-ness, line 27: Can this h1 be outwith the if statement so it doesn't need to be in the else also?

server/middleware/featureFlagMiddleware.js
```diff
@@ -1,5 +1,13 @@
+const config = require('../config')
+
 module.exports = async (req, res, next) => {
-  res.locals.featureFlags = {}
+  res.locals.featureFlags = {
+    three_to_five_policy_change: config.featureFlags.policy_change.three_to_five === 'true',
+  }
+
+  if (req.query?.overrideFeatureFlag !== undefined) {
```
timothy-ness, line 8: It looks like this overrideFeatureFlag is just added in for the integration tests - would it be possible to change the value in config e.g. something like 
const config = require('../../server/config')
config.featureFlags.policy_change.three_to_five = "true"
within your tests and have the feature.env have the feature flag set to false as a default? Rather than intercepting?

server/routes/home.js
```diff
@@ -47,6 +47,14 @@ module.exports = function Index({
 }) {
   const router = express.Router()
 
+  if (process.env.NODE_ENV === 'development') {
```
timothy-ness, line 50: What is this for?

server/middleware/featureFlagMiddleware.js
espedzhev: Yes, I added it for testing purposes. Overriding the feature flag in Cypress is tricky—I initially tried modifying the server config, but it didn’t work for Cypress and didn’t feel like the right approach either. While intercepting isn’t ideal, I like that in development, it allows toggling the feature flag on the go, making testing more flexible. Since it’s going to be removed in a couple of months, I think it should be sufficient for now

server/views/formPages/nextReviewDate/nextReviewDateQuestion.html
espedzhev: That question was raised here: [SI-1449](https://dsdmoj.atlassian.net/browse/SI-1449?focusedCommentId=575979). `conditionalReleaseDate` is not always available, though you mentioned it could just be test data. Maybe there’s a database that can be used to investigate further?

server/views/formPages/nextReviewDate/nextReviewDateStandalone.html
espedzhev: I didn’t notice that. I treated the separation of logic here as ‘the part that will be removed in a couple of months’ and ‘the part that stays.’

server/utils/fieldValidation.js
espedzhev: Yes, you’re correct. The ticket doesn’t mention anything about this. It looks like I had tunnel vision, assuming that everything mentioning 3 years needed to change. I believe the ticket comment asking for clarification also touches on this.

espedzhev: I've rolled back these changes

PR #930, opened 2025-04-01
SI-1511 Move to eslint 9 using HMPPS eslint rules package

espedzhev:
https://dsdmoj.atlassian.net/browse/SI-1511

[instructions](https://github.com/ministryofjustice/hmpps-typescript-lib/blob/main/packages/eslint-config-hmpps/README.md)


eslint.config.mjs
```diff
@@ -0,0 +1,88 @@
...
+
+export default [
+  ...hmppsConfig({
+    extraIgnorePaths: [
+      'integration_tests/**',
+      '**/*.min.js',
+      'integration-tests/build',
+      'node_modules',
+      'eslint.config.mjs',
+    ],
```
espedzhev, line 14:
```suggestion
      'eslint.config.mjs',
      '*.js',
    ],
```

PR #937, opened 2025-04-09
moved config.js to config.ts

espedzhev:


server/services/reviewStatusCalculator.ts
```diff
@@ -1,8 +1,7 @@
 import moment from 'moment/moment'
 
+/* eslint-disable import/prefer-default-export */
```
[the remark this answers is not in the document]
espedzhev: I strongly prefer named exports, and avoiding mixing default + named exports. Named exports also work more reliably when interoperating with CommonJS (require) in TS, since we don’t have to constantly handle .default shims.

It looks like the govuk template follows Airbnb rules (which enables import/prefer-default-export), but we can disable this project-wide if there are no objections @RAGQYFTJ @charlotteianson

server/services/formService.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-shadow */
```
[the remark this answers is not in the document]
espedzhev: There are two config variables in this file (one local, one imported), and I couldn’t get destructuring with aliasing (const { config: conf } = require('../config')) to work correctly in this context. I will try again

server/config.ts
```diff
@@ -0,0 +1,617 @@
...
+  'ESI',
+  'EWI',
+  'BZI',
+  'FHI',
+  'LNI',
+  'SDI',
+  'STI',
+  'NHI',
+  'PFI',
+] as string[]
```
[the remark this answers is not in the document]
espedzhev: Initially, this list was used as a default env value that expected an immutable (readonly) array, and TypeScript would complain about mutability.

Using as string[] was a quick way to satisfy the type checker at the time.

Since we’re now using DEFAULT_FEMALE_PRISON_IDS_ENV instead, this can likely be cleaned up

server/config.ts
```diff
@@ -0,0 +1,631 @@
...
+  PRODUCT_ID: 'UNASSIGNED',
+} as const
+
+function get<T extends FallbackValue>({ name, fallback, log, requireInProduction }: GetParams<T>): T | string {
+  const value = process.env[name]
+
+  if (value !== undefined) {
+    if (log) {
+      logger.info(`Env var: ${name} value: ${value}`)
+    }
```
charlotteianson, line 107: You could tidy this up by using `log && logger.info(...message)` instead of the if statement

server/config.ts
```diff
@@ -0,0 +1,631 @@
...
+
+    return value
+  }
+
+  if (fallback !== undefined && (!production || !requireInProduction)) {
+    if (log) {
+      logger.info(`Env var: ${name} value: ${fallback} (from fallback)`)
```
charlotteianson, line 114: Same as above

server/config.ts
```diff
@@ -0,0 +1,617 @@
...
+
+  // App insights
+  APPLICATIONINSIGHTS_CONNECTION_STRING: '',
+  PRODUCT_ID: 'UNASSIGNED',
+} as const
+
+function get<T extends FallbackValue>({ name, fallback, log, requireInProduction }: GetParams<T>): T | string {
+  const value = process.env[name]
+
+  if (value !== undefined) {
```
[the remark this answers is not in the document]
espedzhev: typeof value !== 'undefined' is useful when checking global or potentially undeclared variables. Since value is already declared here, value !== undefined should be correct

server/config.ts
```diff
@@ -0,0 +1,617 @@
...
+  name: string
+  fallback: T
+  log: boolean
+  requireInProduction: boolean
+}
+
+const DEFAULT_TIMEOUTS = {
+  response: 30000,
+  deadline: 35000,
+} as const
```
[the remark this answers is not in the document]
espedzhev: as const freezes the value — so response becomes the literal type 30000 instead of just number. And it also makes the object immutable.

An interface defines a reusable structure. It’s useful when we want to apply the same shape in multiple places, but it doesn’t preserve literal values or immutability by itself

integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/EventSpecification.groovy
```diff
@@ -201,12 +201,11 @@ class EventSpecification extends AbstractSpecification {
 
     when: 'the job runs'
 
-    def file = new File("../jobs/dead-letter/clearDeadLetterQueuesWithExit.js")
-    def command = "node " + file.absolutePath
+//    def file = new File("../jobs/dead-letter/clearDeadLetterQueuesWithExit.js")
```
timothy-ness, line 204: Are you meaning to keep this in?

timothy-ness (approved):

integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/EventSpecification.groovy
espedzhev: For reference, yes. It could probably be changed to a comment

espedzhev, line 204:
```suggestion
// Reference: ../jobs/dead-letter/clearDeadLetterQueuesWithExit.js
```

integration-tests/src/test/groovy/uk/gov/justice/digital/hmpps/cattool/specs/EventSpecification.groovy
```diff
@@ -201,12 +201,11 @@ class EventSpecification extends AbstractSpecification {
 
     when: 'the job runs'
 
-    def file = new File("../jobs/dead-letter/clearDeadLetterQueuesWithExit.js")
-    def command = "node " + file.absolutePath
+    // Reference: ../jobs/dead-letter/clearDeadLetterQueuesWithExit.js
```
espedzhev, line 204:
```suggestion
    // Reference: ../jobs/dead-letter/clearDeadLetterQueuesWithExit.ts
```

timothy-ness (approved):

PR #940, opened 2025-04-24
Filtering remand prisoners out of the dashboard on a different view of the dashboard

timothy-ness:
In order to test that we can filter out all legal status = REMAND prisoners I've created a new endpoint which shows the exact same as the other endpoint but without the REMAND prisoners, we can use this to also test future changes e.g. the recalls etc.

### DIFF
--- a/server/routes/home.js
+++ b/server/routes/home.js
@@ -205,58 +205,69 @@ module.exports = function Index({
   router.get(
     '/recategoriserHome',
     asyncMiddleware(async (req, res) => {
-      const user = await userService.getUser(res.locals)
-      res.locals.user = { ...user, ...res.locals.user }
-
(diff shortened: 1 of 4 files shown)


server/routes/home.js
```diff
@@ -205,58 +205,69 @@ module.exports = function Index({
...
-      if (validation.error) {
-        logger.error('Recategoriser home page submitted with invalid filters.', validation.error)
-        return res.render('pages/error', {
-          message: 'Invalid recategoriser home filters',
-        })
-      }
+  router.get(
+    '/recategoriserHomeDevelopment',
+    asyncMiddleware(async (req, res) => {
+      return recategoriserHome(req, res, true)
```
espedzhev, line 215: Not a fun of hanging booleans

espedzhev (approved):
LGTM

PR #949, opened 2025-05-16
SI-1482 Recalls issue fix

RAGQYFTJ:


server/services/filter/homeFilter.ts
```diff
@@ -241,7 +242,11 @@ export const filterListOfPrisoners = async (
           }
           break
         case OVERDUE:
-          if (!isReviewOverdue(prisoner.nextReviewDate)) {
+          var nextReviewDate: string = prisoner.nextReviewDate
```
espedzhev, line 245: don't use var please

assets/js/html5shiv-3.7.3.min.js
```diff
@@ -1,4 +1,123 @@
...
\ No newline at end of file
+ * @preserve HTML5 Shiv 3.7.3 | @afarkas @jdalton @jon_neal @rem | MIT/GPL2 Licensed
+ */
```
espedzhev, line 3: this is no longer a min file

espedzhev (changes requested):

server/services/offendersService.js
```diff
@@ -749,9 +748,11 @@ module.exports = function createOffendersService(
       getPomMap(allOffenders, allocationClient),
     ])
 
+    const filteredOutRecalledPrisoners = filterOutRecalledPrisoners(allOffenders, prisonerSearchData, nomisClient)
+
```
espedzhev, line 752:
```suggestion
    const filteredOutRecalledPrisoners = await filterOutRecalledPrisoners(allOffenders, prisonerSearchData, nomisClient)

```

server/services/offendersService.js
```diff
@@ -948,9 +956,11 @@ module.exports = function createOffendersService(
       resultsU21.map(s => [s.bookingId, mapPrisonerSearchDtoToRecategorisationPrisonerSearchDto(s)]),
     )
 
+    const filteredOutRecalledPrisoners = filterOutRecalledPrisoners(eliteCategorisationResultsU21, u21map, nomisClient)
+
```
espedzhev, line 960:
```suggestion
    const filteredOutRecalledPrisoners = await filterOutRecalledPrisoners(eliteCategorisationResultsU21, u21map, nomisClient)

```

server/services/filter/recallFilter.ts
```diff
@@ -0,0 +1,63 @@
...
+  prisoners,
+  prisonerSearchData: Map<number, RecategorisationPrisonerSearchDto>,
+  nomisClient,
+) => {
+  return prisoners.filter(prisoner => {
+    const currentPrisonerSearchData = prisonerSearchData.get(prisoner.bookingId)
+
+    if (currentPrisonerSearchData && currentPrisonerSearchData.recall) {
+      // set dates for recalled prisoners
+      setDatesForRecalledPrisoners(nomisClient, prisoner.prisonerNumber, prisoner.bookingId, currentPrisonerSearchData)
```
espedzhev, line 54:
```suggestion
      await setDatesForRecalledPrisoners(nomisClient, prisoner.prisonerNumber, prisoner.bookingId, currentPrisonerSearchData)
```

espedzhev (changes requested):

server/services/filter/homeFilter.ts
RAGQYFTJ: its been removed.

assets/js/html5shiv-3.7.3.min.js
RAGQYFTJ: this was committed in git in 2019 to make cat tool work on IE8. Dont think we can remove or we may need investigation if we are to remove.

server/services/offendersService.js
RAGQYFTJ: done - looks like linting remove this from my files. fixed it.

server/services/filter/recallFilter.ts
RAGQYFTJ: Fixed

espedzhev (changes requested):
revert min file

test/services/statsService.test.js
```diff
@@ -108,11 +108,11 @@ describe('getRecatFromTo', () => {
...
-      [undefined,         6, undefined, undefined,  6],
-      [8,                 7, undefined, undefined, 15],
-      [undefined, undefined,         5,         2,  7],
-      [undefined, undefined, undefined, undefined,  0],
-      [        8,        13,         5,         2, 28],
+      [undefined, 6, undefined, undefined, 6],
+      [8, 7, undefined, undefined, 15],
+      [undefined, undefined, 5, 2, 7],
+      [undefined, undefined, undefined, undefined, 0],
+      [8, 13, 5, 2, 28],
```
espedzhev, line 115: There are no changes here, so it can be reverted

server/services/filter/recallFilter.ts
```diff
@@ -0,0 +1,74 @@
...
+  const prisonPeriodForBookingId = response.prisonPeriod.find(p => p.bookingId === bookingId)
+  if (prisonPeriodForBookingId) {
+    const dateInPrisonSortedDesc = prisonPeriodForBookingId.movementDates.sort(
+      (a, b) => moment(b.dateInToPrison).valueOf() - moment(a.dateInToPrison).valueOf(),
+    )
+    // Record dates for recalled prisoners
+    // eslint-disable-next-line no-param-reassign
+    recatPrisonerSearchDto.dueDateForRecalls = dueDateForRecalls(dateInPrisonSortedDesc[0].dateInToPrison)
+    // eslint-disable-next-line no-param-reassign
+    recatPrisonerSearchDto.lastDateInPrison = moment(dateInPrisonSortedDesc[0].dateInToPrison).format('YYYY-MM-DD')
```
espedzhev, line 20: Should be pure: don’t mutate inputs

server/services/filter/recallFilter.ts
```diff
@@ -0,0 +1,74 @@
...
+        await setDatesForRecalledPrisoners(
+          nomisClient,
+          prisoner.offenderNo || prisoner.prisonerNumber,
+          prisoner.bookingId,
+          currentPrisonerSearchData,
+        )
+
+        if (isFixedTermRecallLessThanAndEqualTo28Days(currentPrisonerSearchData)) {
+          return false // Filter out
+        }
```
espedzhev, line 66: This block currently mutates the original DTO(Data Transfer Object), which makes the side-effects hard to follow. Cleaner would be:

- Fetch the `movementDates` for this booking ID from Nomis
- Compute `lastDateInPrison` and `dueDateForRecalls` from those movementDates
- Build a new DTO (cloning the existing one and adding those dates)
- Use the new DTO to check `isFixedTermRecallLessThanAndEqualTo28Days` and filter accordingly

espedzhev (changes requested):

test/services/statsService.test.js
RAGQYFTJ: fixed

server/services/filter/recallFilter.ts
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      return true
+    }),
+  )
+
+  return prisoners.filter((_, index) => results[index])
```
espedzhev, line 73: Basically, we’re looping through the `prisoners` array and deciding which ones to keep. It can be simplified to:

```ts
const kept: Prisoner[] = []

for (const prisoner of prisoners) {
  let keep = true

  // logic to compute updatedDto

  keep = !isFixedTermRecallLessThanAndEqualTo28Days(updatedDto)

  if (keep) {
    kept.push(prisoner)
  }
}

return kept

**Comment by espedzhev** [comment]
Generally, immutability is the preferred pattern in modern JS/TS. As a rule of thumb, it makes code cleaner, easier to follow, and helps avoid unexpected side effects. While there are cases where mutation is unavoidable, we should favour immutable updates wherever possible

File: assets/js/html5shiv-3.7.3.min.js [hunk 10]
```diff
@@ -1,4 +1,4 @@
...
...re,footer,header,hgroup,main,nav,section{display:block}mark{background:#FF0;color:#000}template{display:none}")),l||i(a,d),a}var k,l,m="3.7.3",n=a.html5||{},o=/^<|^(?:button|map|select|textarea|object|iframe|option|optgroup)$/i,p=/^(?:a|b|code|div|fieldset|h1|h2|h3|h4|h5|h6|i|label|li|ol|p|q|span|strong|style|table|tbody|td|th|tr|ul)$/i,q="_html5shiv",r=0,s={};!function(){try{var a=b.createElement("a");a.innerHTML="<xyz></xyz>",k="hidden"in a,l=1==a.childNodes.length||function(){b.createElement("a");var a=b.createDocumentFragment();return"undefined"==typeof a.cloneNode||"undefined"==typeof a.createDocumentFragment||"undefined"==typeof a.createElement}()}catch(c){k=!0,l=!0}}();var t={elements:n.elements||"abbr article aside audio bdi canvas data datalist details dialog figcaption figure footer header hgroup main mark meter nav output picture progress section summary template time video",version:m,shivCSS:n.shivCSS!==!1,supportsUnknownElements:l,shivMethods:n.shivMethods!==!1,type:"default",shivDocument:j,createElement:g,createDocumentFragment:h,addElements:e};a.html5=t,j(b),"object"==typeof module&&module.exports&&(module.exports=t)}("undefined"!=typeof window?window:this,document);
```
espedzhev, line 4: I would prefer no changes to this file, so the commit history remains intact

server/data/nomis/prisonTimeline/PrisonApiPrisonPeriods.dto.ts
```diff
@@ -0,0 +1,12 @@
+export interface PrisonApiPrisonPeriodsDto {
+  prisonerNumber: string
+  prisonPeriod:
+    | {
+        bookingNumber: string
```
RAGQYFTJ, line 5: do we need bookingNumber or bookingId is just enough. Not really bothered but can remove if not required

RAGQYFTJ: Ok I see you are using it ignore.

espedzhev (approved):
Looks like it’s on the right track to me.

espedzhev (approved):

PR #958, opened 2025-06-20
SI-1671: Load risk of escape data from Alerts API

charlotteianson:
Updates the data source for escape information from risk profiler to the alerts api.

Ticket: [SI-1671](https://dsdmoj.atlassian.net/jira/software/c/projects/SI/boards/1800?assignee=712020%3Ae5c031f1-6420-4bfe-b3dc-d1da4bcd14ce&selectedIssue=SI-1671)

[SI-1671]: https://dsdmoj.atlassian.net/browse/SI-1671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/routes/recat.test.js
```diff
@@ -123,7 +127,7 @@ beforeEach(() => {
   riskProfilerService.getSecurityProfile.mockResolvedValue({})
   riskProfilerService.getViolenceProfile.mockResolvedValue({})
   riskProfilerService.getExtremismProfile.mockResolvedValue({})
-  riskProfilerService.getEscapeProfile.mockResolvedValue({})
+  riskService.getEscapeProfile.mockResolvedValue({})
```
espedzhev, line 130: Have considered `alertService` as alternative name?

espedzhev (dismissed):

test/routes/recat.test.js
charlotteianson: I think initially when i named it `riskService` i thought it would have stuff related to the other apis replacing risk profiler too but now i think it's best to separate them so `alertService` is probs better 😄 i'll change it

espedzhev (approved):

integration_tests/mockApis/alertsApi.ts
```diff
@@ -0,0 +1,80 @@
...
+              code: alertCode,
+            },
+            activeFrom: '2016-09-14',
+          },
+        ],
+      },
+    },
+  })
+
+const stubGetProfileWomenEscapeAlert = ({
```
timothy-ness, line 32: Is this the exact same as the function above?

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "offender-categorisation",
   "version": "0.0.1",
   "scripts": {
-    "start": "node dist/server.js | bunyan -o short",
+    "start": "export $(cat .env.groovy) && node dist/server.js | bunyan -o short",
```
timothy-ness, line 5: Don't think this needs to be here anymore

server/data/alertsApi/alertsApiClient.test.ts
```diff
@@ -0,0 +1,44 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import nock from 'nock'
+import { config } from '../../config'
+import { makeTestUser } from '../user.test-factory'
+import { alertsApiClientBuilder } from './alertsApiClient'
+import { makeTestEscapeAlertDto } from './escapeAlert.dto.test-factory'
+
+jest.mock('redis', () => ({
```
timothy-ness, line 8: Doesn't look like you're using redis within the API client so I don't think you need this?

server/utils/alertServiceHelpers.ts
```diff
@@ -0,0 +1,46 @@
+import {
```
timothy-ness, line 1: I'd probably call this something like escapeProfileMapper (and then name the method in it mapAlertToEscapeProfile) or alertTransformer

integration_tests/mockApis/alertsApi.ts
charlotteianson: It is the same, that's how it was done before inside the risk profiler mock api - i did notice this when i was working on it but just left it as it was to differentiate in the women's estate tests, guess i could just use the same one though and save on code

server/data/alertsApi/alertsApiClient.test.ts
charlotteianson: I thought that but then when i took it out and ran the tests without it they failed, think cause the api client uses `getApiClientToken`

integration_tests/mockApis/alertsApi.ts
charlotteianson: I've removed it and just used the same one 😄

PR #972, opened 2025-07-28
Recalls due date should be 10 calendar days after recall

timothy-ness:
It was 10 business days (so often 14 days rather than 10), but with the policy it states it should just be 10 days after recall (e.g. due recalled date + 10 days, overdue on day 11)


server/services/offendersService.js
```diff
@@ -808,8 +810,10 @@ module.exports = function createOffendersService(
           ) {
             return null
           }
-          // Recalls are due a recategorisation within 10 business days of the recall date rather then the original next review date
-          reviewDueDate = add10BusinessDays(recalledOffenderData.get(nomisRecord.offenderNo).recallDate)
+          // Recalls are due a recategorisation within 10 days of the recall date rather then the original next review date
```
espedzhev, line 813: It will be worth to say its exclusive with an a example eg `if the recall is the 1st, the establishment has 10 days to do a cat review before it becomes overdue so if the review has not been completed by the end of the 10th it should become overdue on the 11th`

espedzhev (dismissed):

PR #978, opened 2025-08-08
SI-1736 Update OAuth client to get rid of refresh token - [CAT]

espedzhev:
[Update OAuth client to get rid of refresh token - [CAT]](https://dsdmoj.atlassian.net/browse/SI-1736)


server/app.js
charlotteianson, line 1: Just checking all the comments in this file were left in intentionally?

espedzhev: yeah, I copied those from manage soc cases

PR #1047, opened 2026-06-05
use getLatestActiveCategorisationForOffenders for security referrals

charlotteianson:
[Ticket](https://dsdmoj.atlassian.net/jira/software/c/projects/SI/boards/1800?selectedIssue=SI-2337)

This pull request introduces a new method to fetch only the latest active categorisations for offenders, updates the offenders service to use this method for recategorisation security referrals, and ensures the corresponding tests are updated and expanded. The main goal is to improve data accuracy by filtering for active categorisations in relevant scenarios.

### DIFF
--- a/server/services/offendersService.js
+++ b/server/services/offendersService.js
@@ -368,7 +368,7 @@ module.exports = function createOffendersService(
         const [offenderDetailsFromNomis, userDetailFromElite, nomisCatData] = await Promise.all([
           nomisClient.getOffenderDetailList(securityReferredFromDB.map(c => c.offenderNo)),
           nomisClient.getUserDetailList(securityReferredFromDB.map(c => c.securityReferredBy)),
-          nomisClient.getLatestCategorisationForOffenders(
+          nomisClient.getLatestActiveCategorisationForOffenders(
             securityReferredFromDB.filter(c => c.catType === CatType.RECAT.name).map(c => c.offenderNo),
           ),
         ])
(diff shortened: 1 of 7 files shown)


server/services/offendersService.js
```diff
@@ -368,7 +368,7 @@ module.exports = function createOffendersService(
         const [offenderDetailsFromNomis, userDetailFromElite, nomisCatData] = await Promise.all([
           nomisClient.getOffenderDetailList(securityReferredFromDB.map(c => c.offenderNo)),
           nomisClient.getUserDetailList(securityReferredFromDB.map(c => c.securityReferredBy)),
-          nomisClient.getLatestCategorisationForOffenders(
+          nomisClient.getLatestActiveCategorisationForOffenders(
```
a6software, line 371: there's another call to this in this file (line 492 / line 510) - just checking that one doesn't also need to change? As in, do we not always want the latest active? (I don't know, just asking)

charlotteianson: I deliberately only changed that one call to keep the fix narrow but agree it's probs worth checking the other usages, i'll have a look this morn

charlotteianson: @a6software so i looked into this and replaced one usage as it was only using the endpoint for `nextReviewDate`, so pretty confident that should also be using the last active/approved categorisation.

The other usage (inside `mergeU21ResultWithNomisCategorisationData`) looks like it intentionally includes pending categorisations and only filters out inactive ones:

```
const eliteResultsRaw = await nomisClient.getLatestCategorisationForOffenders(
      resultsU21IJ.map(c => c.prisonerNumber),
    )

// results can include inactive - need to remove
const eliteResultsFiltered = eliteResultsRaw.filter(c => c.assessmentStatus !== 'I')
```

That one is using `assessmentStatus` rather than `nextReviewDate` so don't think we should change it to active only as would also exclude pending categorisations and potenitally change the intended behaviour

a6software (approved):

PR #1051, opened 2026-06-29
SI-2287 Upgrade express to v5

kasiawalshw:
Upgrade Helmet
upgrade express to v5 and types/express
add setupRequestParsing
remove body-parser as per template
remove asyncMiddleware
refactor landing routes in home.js as express 5 doesn't like regexp

Ticket: https://dsdmoj.atlassian.net/browse/SI-2287


charlotteianson (approved):
LGTM

PR #1052, opened 2026-07-03
Pnomis when lite cat

timothy-ness:
This ticket is to check if the nominal showing in the categorisation dashboard is U because they have had a lite cat, meaning that the due date should be the next review date of the lite cat rather than 10 days after the sentence. Right now they are showing with 'PNOMIS' next to them which is a bug.


server/services/offendersService.js
```diff
@@ -247,19 +247,15 @@ module.exports = function createOffendersService(
...
           const pnomis = liteInProgress
             ? 'OTHER'
-            : (inconsistent || (nomisStatusAwaitingApproval && !dbRecord.status)) && 'PNOMIS'
-
-          if (pnomis === 'PNOMIS') {
-            logger.warn(
-              `getUncategorisedOffenders: Marking record as PNOMIS for booking id=${nomisRecord.bookingId}, offenderNo=${nomisRecord.offenderNo} due to status inconsistency or missing local record, Nomis status=${nomisRecord.status}, PG status=${dbRecord.status}`,
-            )
-          }
+            : (!assessmentData.category || assessmentData.category !== 'U') &&
```
espedzhev, line 250:
```suggestion
            : assessmentData.category !== 'U' &&
```

server/services/offendersService.js
```diff
@@ -1175,15 +1183,18 @@ module.exports = function createOffendersService(
...
     }
     const sentenceDateMoment = moment(sentenceDate, 'YYYY-MM-DD')
     const daysSinceSentence = moment().diff(sentenceDateMoment, 'days')
 
     const actualDays = get10BusinessDays(sentenceDateMoment)
-    const dateRequiredRaw = sentenceDateMoment.add(actualDays, 'day')
+    const dateRequiredRaw =
+      typeof liteCat === 'undefined' || !liteCat.approvedDate
+        ? sentenceDateMoment.add(actualDays, 'day')
+        : moment(liteCat.nextReviewDate, 'YYYY-MM-DD')
```
espedzhev, line 1197: if `liteCat.nextReviewDate` is `null` it will return `Moment<Invalid date>`

server/services/offendersService.js
```diff
@@ -379,40 +375,43 @@ module.exports = function createOffendersService(
...
-            const entry = sentenceMap.get(o.bookingId)
-            sentenceAndDate = entry && buildSentenceData(entry.sentenceDate)
-          } else {
-            const nomisCat = nomisCatData.find(record => record.bookingId === o.bookingId)
-            sentenceAndDate = { dateRequired: nomisCat && dateConverter(nomisCat.nextReviewDate) }
-          }
+            let sentenceAndDate
+            if (o.catType === CatType.INITIAL.name) {
+              const entry = sentenceMap.get(o.bookingId)
+              const assessmentData = await formService.getLiteCategorisation(o.bookingId)
```
espedzhev, line 397: This is an N+1 query (one DB call per offender). Given our small services it's probably fine for now, but worth watching if performance degrades. It could be avoided by fetching all booking IDs in one go and mapping the results, but that's extra effort. Your call

a6software (approved):
honestly pretty tricky to review - I can't see any issues with the change

but the 2000 line file needs a serious rethink

PR #1053, opened 2026-07-07
SI-2404: Move applicationinsights to @ministryofjustice/hmpps-azure-telemetry

charlotteianson:
This pull request migrates the application's telemetry and monitoring from the legacy applicationinsights package to the new @ministryofjustice/hmpps-azure-telemetry library. It updates configuration, environment variable usage, and initialization logic to support this new approach, and adds comprehensive tests for the telemetry setup.


helm_deploy/offender-categorisation/values.yaml
```diff
@@ -39,36 +39,35 @@ generic-service:
     failureThreshold: 15
 
   env:
-    DB_SSL_ENABLED: "true"
-    REDIS_TLS_ENABLED: "true"
-    APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/"
```
espedzhev, line 44: is this an intentional change?

package.json
```diff
@@ -89,9 +89,9 @@
     "@microsoft/applicationinsights-clickanalytics-js": "^3.1.0",
     "@microsoft/applicationinsights-web": "^3.1.0",
     "@ministryofjustice/frontend": "^3.7.0",
+    "@ministryofjustice/hmpps-azure-telemetry": "1.0.2",
```
espedzhev:
```suggestion
    "@ministryofjustice/hmpps-azure-telemetry": "^1.0.2",
```

espedzhev (dismissed):

helm_deploy/offender-categorisation/values.yaml
charlotteianson: Yeah - i'm assuming the additional endpoints aren't required cause the other apps that have already been migrated all use the shorter connection string format and telemetry works fine in those...

Happy to switch back if you think differently though?

timothy-ness (dismissed):

helm_deploy/offender-categorisation/values.yaml
espedzhev: Maybe @timothy-ness knows? https://github.com/ministryofjustice/offender-categorisation/pull/848

charlotteianson: @espedzhev i've tested it in dev so fairly confident it's all working as it should do

espedzhev (approved):

espedzhev:
The console shows a CSP violation blocking telemetry POSTs to  `dc.services.visualstudio.com`
Screenshot 2026-07-07 at 17 16 52

This is caused by simplifying  `APPLICATIONINSIGHTS_CONNECTION_STRING`  in this PR's  `values.yaml`. That same env var also feeds the client-side App Insights snippet in `server/views/partials/layout.html`.

Previously the connection string explicitly pointed the Web SDK at `https://northeurope-0.in.applicationinsights.azure.com`, which matches the CSP `connectSrc` allowlist in `setUpWebSecurity.js`. 

Without an explicit `IngestionEndpoint`, the SDK falls back to its default endpoint,  `dc.services.visualstudio.com`, which isn't whitelisted, so the browser blocks the POST.

There are two options to fix it:

1. Keep an explicit IngestionEndpoint (and  LiveEndpoint  if needed) in the connection string, matching the new telemetry setup
2. Add `https://dc.services.visualstudio.com` to `connectSrc` in `server/utils/setUpWebSecurity.js` to allow the default endpoint:

```ts
const azureUrls = [
  'https://northeurope-0.in.applicationinsights.azure.com',
  '*.monitor.azure.com',
  'https://dc.services.visualstudio.com',
]
```

Option 2 requires confirmation from @timothy-ness

espedzhev (changes requested):
Need to investigate CSP console error

charlotteianson:
> The console shows a CSP violation blocking telemetry POSTs to  `dc.services.visualstudio.com` Screenshot 2026-07-07 at 17 16 52
> 
> This is caused by simplifying  `APPLICATIONINSIGHTS_CONNECTION_STRING`  in this PR's  `values.yaml`. That same env var also feeds the client-side App Insights snippet in `server/views/partials/layout.html`.
> 
> Previously the connection string explicitly pointed the Web SDK at `https://northeurope-0.in.applicationinsights.azure.com`, which matches the CSP `connectSrc` allowlist in `setUpWebSecurity.js`.
> 
> Without an explicit `IngestionEndpoint`, the SDK falls back to its default endpoint,  `dc.services.visualstudio.com`, which isn't whitelisted, so the browser blocks the POST.
> 
> There are two options to fix it:
> 
> 1. Keep an explicit IngestionEndpoint (and  LiveEndpoint  if needed) in the connection string, matching the new telemetry setup
> 2. Add `https://dc.services.visualstudio.com` to `connectSrc` in `server/utils/setUpWebSecurity.js` to allow the default endpoint:
> 
> ```ts
> const azureUrls = [
>   'https://northeurope-0.in.applicationinsights.azure.com',
>   '*.monitor.azure.com',
>   'https://dc.services.visualstudio.com',
> ]
> ```
> 
> Option 2 requires confirmation from timothy-ness

Good spot, thanks for looking into this... i've just changed it back now

espedzhev (approved):

